PR #10, opened 2024-02-19
New Plugin Spans

TheMutsi:
Expose Span methods to RN (recordSpanWithName is a custom method in JavaScript to manage the async functions)

Methods
```
startSpanWithName: (name: string, parentSpanId?: string) => Promise<boolean | string>;

stopSpanWithId: (spanId: string, errorCode?: SPAN_ERROR_CODES) => Promise<boolean>;

addSpanEventToSpanId: (spanId: string, name: string, timestamp?: number, attributes?: Attributes) => Promise<boolean>;

addSpanAttributesToSpanId: (spanId: string, key: string, value: string) => Promise<boolean>;

recordSpanWithName: (name: string, callback: () => void | Promise<void>, attributes?: Attributes, events?: Events[], parentSpanId?: string) => Promise<boolean>;

recordCompletedSpanWithName: (name: string, startTimestamp: number, endTimestamp: number, errorCode?: SPAN_ERROR_CODES, parentSpanId?: string, attributes?: Attributes, events?: Events[]) => Promise<boolean>;
```

Types
```
interface Attributes {
  [key: string]: string;
}
interface Events {
  name: string;
  timestamp?: number;
  attributes?: Attributes;
}

type SPAN_ERROR_CODES = 'None' | 'Failure' | 'UserAbandon' | 'Unknown';
```

Tests
Screenshot 2024-03-08 at 5 25 30 PM

Session Android (Spans + Breadcrumb + Network + Moments + Logs)
https://dash.embrace.io/app/p89ke/grouped_sessions/last_1_hour/F64F0004A27842C8AAB33E9A55DC3F91/2DED083F1C0B4EF9977B6D7AB54FE930@/2DED083F1C0B4EF9977B6D7AB54FE930?first=08B88F8B133344F6A4FA50D235AA33D4&last=2DED083F1C0B4EF9977B6D7AB54FE930

Session iOS (Spans + Breadcrumb + Network + Moments + Logs)
https://dash.embrace.io/app/A82U7/grouped_sessions/last_3_hour/17C1830B8BB443EE9B440703F2B5149C/56E0CD5FDB75479F9061FA6C68C75DCD@/56E0CD5FDB75479F9061FA6C68C75DCD?first=7B2E15C9490140D1B8215C43A66C9D75&last=56E0CD5FDB75479F9061FA6C68C75DCD


.github/CODEOWNERS
```diff
@@ -0,0 +1,5 @@
+# Global owners
+* @embrace-io/team-react-native
```
bidetofevil, line 2: It says this team is not public or doesn't have write access? Is it because you haven't created it yet?

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,31 @@
...
+  build:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [16.x]
+        # See supported Node.js release schedule at https://nodejs.org/en/about/releases/
+
+    steps:
+      - uses: actions/checkout@v3
```
bidetofevil, line 18: Android is using v4 now so it's probably safe for you to upgrade too

packages/action-tracker/README.md
```diff
@@ -28,8 +28,6 @@ npm install @embrace-io/action-tracker
 
 ### Adding the component to your code
 
-Add an useRef for the NavigationContainer and pass it to Embrace's hook
```
bidetofevil, line 31: This is just old code you're deleting?

packages/core/android/gradle.properties
```diff
@@ -1 +1 @@
-emb_android_sdk=6.4.0
```
bidetofevil, line 1: Whitespace change? Not sure sure if it's matters

packages/core/android/src/main/java/io/embrace/embracewrapper/EmbraceManagerModule.java
```diff
@@ -562,6 +568,110 @@ public void logNetworkClientError(String url,
...
+        List<Map<String, Object>> objectMapList = new ArrayList<>();
+
+        try{
+            for (int i = 0; i < items.size(); i++) {
+                ReadableMap readableMap = items.getMap(i);
+
+                if (readableMap != null) {
+                    Map<String, Object> map = readableMap.toHashMap();
+                    // TODO Change when Android/iOS replace time in nano for ms
+                    if (map.containsKey("timestampNanos") && map.get("timestampNanos") instanceof Double) {
```
bidetofevil, line 620: So Android should not hit this as Android returns a Long if that field is populated?

packages/core/android/src/main/java/io/embrace/embracewrapper/EmbraceManagerModule.java
```diff
@@ -562,6 +568,110 @@ public void logNetworkClientError(String url,
...
+            
+            ErrorCode errorCodeInstance = this.getSpanErrorCodebyString(errorCodeString);
+            promise.resolve(Embrace.getInstance().getReactNativeInternalInterface().stopSpan(spanId, errorCodeInstance));
+        }catch(Exception e){
+            promise.resolve(false);
+        }
+    }
+
+    @ReactMethod()
+    public void addSpanAttributesToSpanId(String spanId, String key, String value, Promise promise) {
```
bidetofevil, line 646: So you're not adding to the spanId but to the span with the spanId.

Does it make sense to call this `addSpanAttributeToSpan()` or `addSpanAttributeWithSpanId()`?

Also it's singular - attribute not attributes 😅

packages/core/android/src/main/java/io/embrace/embracewrapper/EmbraceManagerModule.java
```diff
@@ -562,6 +568,110 @@ public void logNetworkClientError(String url,
...
+        try{
+            promise.resolve(Embrace.getInstance().getReactNativeInternalInterface().addSpanAttribute(spanId, key, value));
+        }catch(Exception e){
+            promise.resolve(false);
+        }
+    }
+
+    @ReactMethod()
+    public void addSpanEventToSpanId(String spanId, String name, Double time, ReadableMap attributes, Promise promise) {
```
bidetofevil, line 655: Same comment as above with the method name, re: ToSpan vs WithSpanId

packages/core/android/src/main/java/io/embrace/embracewrapper/EmbraceManagerModule.java
```diff
@@ -562,6 +568,110 @@ public void logNetworkClientError(String url,
...
+        try{
+
+            promise.resolve(Embrace.getInstance().getReactNativeInternalInterface().addSpanEvent(spanId, name, time.longValue(), this.convertToReadableMap(attributes)));
+        }catch(Exception e){
+            promise.resolve(false);
+        }
+    }
+
+    @ReactMethod()
+    public void recordCompletedSpanWithName(String name, Double startTimeNanos, Double endTimeNanos, String errorCodeString, String parentSpanId, ReadableMap attributes, ReadableArray events, Promise promise) {
```
bidetofevil, line 665: Lets use millis in the public API so we never get nanos coming from RN at all

packages/core/android/src/main/java/io/embrace/embracewrapper/EmbraceManagerModule.java
```diff
@@ -562,6 +568,110 @@ public void logNetworkClientError(String url,
...
+        try{
+            ErrorCode errorCodeInstance = this.getSpanErrorCodebyString(errorCodeString);
+
+            promise.resolve(Embrace.getInstance().getReactNativeInternalInterface().recordCompletedSpan(name, startTimeNanos.longValue(), endTimeNanos.longValue(), errorCodeInstance, parentSpanId, this.convertToReadableMap(attributes), this.transformListReadableMapToListMap(events)));
```
bidetofevil, line 669: This method expects millis now starting with 6.4

packages/spans/src/index.ts
```diff
@@ -0,0 +1,199 @@
...
+      e
+    );
+    return createFalsePromise();
+  }
+};
+
+export const addSpanEventToSpanId = (
+  spanId: string,
+  name: string,
+  timeStamp?: number,
```
bidetofevil, line 48: Do we usually add the unit (e.g. ms) to parameter names by convention? If it's ok, timeStampMs is preferred since the OTel JavaScript SDK uses nanos and we don't, so making that clear would be nice

This applies to all timestamps BTW.

packages/spans/src/index.ts
```diff
@@ -0,0 +1,199 @@
...
+import { Attributes, Events, SPAN_ERROR_CODES } from '../interfaces/ISpans';
+import {
+  convertMSToNano,
+  createFalsePromise,
+  validateAndLogRequiredProperties,
+} from './Utils';
+
+export const startSpanWithName = (
+  name: string,
+  parentSpanId?: string
```
bidetofevil, line 11: if you can do a default parameter to null here, that would be useful - most folks will be creating parent-less spans to start off with so not making them pass a null would be great.

or does TypeScript still do that for you if you drop trailing parameters still? JS did and I hated it so I was hoping TS would've removed that lol

bidetofevil (commented):
I looked at the Java and JS code only (not iOS) and it mostly looks good.

The one thing that should change is the public API should expose milliseconds when it comes to timestamps (ideally it would be in the name of the parameter as well). Some other comments about method naming would be nice to do but it's up to you.

I'll approve it from my end once the time units are fixed for the API

packages/core/android/src/main/java/io/embrace/embracewrapper/EmbraceManagerModule.java
TheMutsi: I implemented iOS first and they call it that way. should I change ? 
image

packages/spans/src/index.ts
TheMutsi: JS and TS will pass it as undefined (similar to null)

packages/core/android/src/main/java/io/embrace/embracewrapper/EmbraceManagerModule.java
TheMutsi: Note: We agreed to use Nanos until iOS change it

packages/core/android/src/main/java/io/embrace/embracewrapper/EmbraceManagerModule.java
TheMutsi: I dont have Long in my "types whitelist"
image
2 lines bellow I cast it to long

packages/action-tracker/README.md
TheMutsi: Yes we dont use it anymore

packages/core/android/src/main/java/io/embrace/embracewrapper/EmbraceManagerModule.java
```diff
@@ -562,6 +568,110 @@ public void logNetworkClientError(String url,
...
+        try{
+
+            promise.resolve(Embrace.getInstance().getReactNativeInternalInterface().addSpanEvent(spanId, name, time.longValue(), this.convertToReadableMap(attributes)));
+        }catch(Exception e){
+            promise.resolve(false);
+        }
+    }
+
+    @ReactMethod()
+    public void recordCompletedSpan(String name, Double startTimeNanos, Double endTimeNanos, String errorCodeString, String parentSpanId, ReadableMap attributes, ReadableArray events, Promise promise) {
```
bidetofevil, line 665: Since this is open sourced, you might want to add a comment here saying the Android SDK expects millisecond but this works for now because of the workaround.

bidetofevil (approved):
Looks good from the Android side!

packages/core/android/src/main/java/io/embrace/embracewrapper/EmbraceManagerModule.java
TheMutsi: I will let it with that signature, since I need to match the parameters with iOS. In the next version , when iOS update the parameter, I will refactor this

PR #27, opened 2024-06-05
Change the default location of the sourcemap, since we were asking to…

TheMutsi:
## Goal

Change the default location of the sourcemap, since we were asking to store it inside the app file

https://www.notion.so/embraceio/RN-IOS-SourceMap-inside-IPA-66d4d11e45eb4deb9457037b6ccf6c17

### DIFF
--- a/packages/core/scripts/setup/ios.ts
+++ b/packages/core/scripts/setup/ios.ts
@@ -144,7 +144,8 @@ export const addUploadBuildPhase = {
             null,
             {
               shellPath: '/bin/sh',
-              shellScript: `EMBRACE_ID=${id} EMBRACE_TOKEN=${token} ${embRunScript}`,
+              shellScript: `export REACT_NATIVE_MAP_PATH="$CONFIGURATION_BUILD_DIR/main.jsbundle.map"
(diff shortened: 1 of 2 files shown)


packages/core/scripts/setup/ios.ts
```diff
@@ -144,7 +144,8 @@ export const addUploadBuildPhase = {
             null,
             {
               shellPath: '/bin/sh',
-              shellScript: `EMBRACE_ID=${id} EMBRACE_TOKEN=${token} ${embRunScript}`,
+              shellScript: `export REACT_NATIVE_MAP_PATH="$CONFIGURATION_BUILD_DIR/main.jsbundle.map"
```
jpmunz, line 147: is adding `export` needed here or are the env vars only relevant in the context of `embRunScript`?

jpmunz (approved):

packages/core/scripts/setup/ios.ts
TheMutsi: The variable is not being used in that scope, but inside run.sh. You need the export to be used inside the run.sh

PR #28, opened 2024-06-07
[EMBR-3872] Instrumentation library(ies) for RN navigation

facostaembrace:
## Goal

Introduce new instrumentation library that creates telemetry data under "experimental" mode that tracks changes into the navigation.

Pending:
- isolate package by moving related dependencies into the package itself.
- decide the approach to create a bundle


packages/experimental/navigation/README.md
```diff
@@ -0,0 +1,4 @@
+## OTel Instrumentation Library - React Native / Navigation
```
jpmunz, line 1: could add a few usage examples as well similar to something like https://github.com/open-telemetry/opentelemetry-js/tree/main/experimental/packages/opentelemetry-instrumentation-http to show how the different instrumentations would be used in an app

given we're introducing the "experimental" folder for the first time I think a quick `packages/experimental/README.md` would be worthwhile as well, we can chat on slack about what should go in there

packages/experimental/navigation/components/NativeNavigationTracker.tsx
```diff
@@ -0,0 +1,37 @@
...
+  /**
+   * NOTE
+   * - should we check if the reference is the one we need even when typescript would complain if it is different?
+   * (People can force the type deactivating the typescript checks)
+   * - should introduce a log to warn the user if the ref is not the one we are expecting and return?
+   */
+
+  useNativeNavigationTracker(ref, tracer);
+
+  return <>{children}</>;
```
jpmunz, line 32: would just `return children` be equivalent?

packages/experimental/navigation/hooks/useNativeNavigationTracker.ts
```diff
@@ -0,0 +1,62 @@
...
+      if (!componentName) {
+        // do nothing in case for some reason there is no route
+        return;
+      }
+
+      spanCreator(tracer, span, navView, componentName);
+    });
+
+    navigationElRef.registerCommandListener(() => {
+      // NOTE: To implement
```
jpmunz, line 55: what would we be looking to listen for beyond the element appear/disappear?

packages/experimental/navigation/otel/hooks/useTrace.ts
```diff
@@ -0,0 +1,39 @@
...
+
+const useTrace = (
+  {name, version}: ConfigArgs = {},
+  provider: TracerProvider,
+): TracerRef => {
+  // otel
+  const tracerRef = useRef<Tracer | null>(null);
+
+  // using the layout effect to make sure the tracer is initialized before the component is rendered
+  useEffect(() => {
```
jpmunz, line 24: given the comment was this meant to be using `useLayoutEffect`?

packages/experimental/navigation/otel/hooks/useTrace.ts
```diff
@@ -0,0 +1,39 @@
...
+    if (tracerRef.current === null && provider) {
+      trace.setGlobalTracerProvider(provider);
```
jpmunz, line 26: I think we should avoid this in the library since it may have an unintended effect on the rest of the app's instrumentation, should be able to do `provider.getTracer(...)` without setting it to be global

packages/experimental/navigation/otel/hooks/useTrace.ts
```diff
@@ -0,0 +1,39 @@
...
+};
+
+interface ConfigArgs {
+  name?: string;
+  version?: string;
+}
+
+type TracerRef = MutableRefObject<Tracer | null>;
+
+const useTrace = (
```
jpmunz, line 16: minor, I think `useTracer` would be slightly more consistent with the other naming

packages/experimental/navigation/otel/hooks/useTrace.ts
```diff
@@ -0,0 +1,39 @@
...
+  // otel
+  const tracerRef = useRef<Tracer | null>(null);
+
+  // using the layout effect to make sure the tracer is initialized before the component is rendered
+  useEffect(() => {
+    if (tracerRef.current === null && provider) {
+      trace.setGlobalTracerProvider(provider);
+
+      tracerRef.current = trace.getTracer(
+        name || TRACER_DEFAULT.name,
```
jpmunz, line 29: given 'name' is required in `getTracer` I'd maybe make that required in this hook as well and then maybe just omit a version when it isn't set since '1.0' may not always be reasonable as a default

packages/experimental/navigation/package.json
jpmunz, line 1: we may need to look at our package.json files as a separate task to make sure they are consistent, noted a couple quick changes for now

packages/experimental/navigation/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "navigation",
+  "version": "1.0.0",
```
jpmunz, line 3: maybe "0.1.0" since we're keeping under experimental for now

packages/experimental/navigation/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "description": "Instrumentation library that captures telemetry data during navigation changes.",
+  "author": "Embrace.io",
+  "license": "UNLICENSED",
```
jpmunz, line 6: Apache-2.0 to match the top level

packages/experimental/navigation/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "main": "index.ts",
+  "scripts": {
+    "test": "jest"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git@github.com/embrace-io/embrace-io.git"
```
jpmunz, line 13: https://github.com/embrace-io/embrace-react-native-sdk.git

packages/experimental/navigation/package.json
```diff
@@ -0,0 +1,44 @@
...
+  },
+  "keywords": [
+    "otel",
+    "react-native",
+    "navigation",
+    "telemetry",
+    "instrumentation"
+  ],
+  "bugs": {
+    "url": "https://github.com/embrace-io/embrace-io/issues"
```
jpmunz, line 41: https://github.com/embrace-io/embrace-react-native-sdk/issues

packages/experimental/navigation/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "@testing-library/react": "^16.0.0",
+    "@testing-library/react-native": "^12.5.1",
+    "@types/jest": "^29.5.12",
+    "@types/react": "~18.2.45",
+    "@types/react-dom": "^18.3.0",
+    "@types/react-native": "^0.73.0",
+    "@types/react-test-renderer": "^18.0.7"
+  },
+  "keywords": [
+    "otel",
```
jpmunz, line 34: I'd add "opentelemetry" spelled out as well, e.g. https://github.com/open-telemetry/opentelemetry-js/blob/main/experimental/packages/exporter-logs-otlp-http/package.json#L51

packages/experimental/navigation/components/NavigationTracker.tsx
```diff
@@ -0,0 +1,36 @@
...
+  // selected provider, should be configured by the app consumer
+  provider: TracerProvider;
+};
+
+const NavigationTracker = forwardRef<
+  NavigationTrackerRef,
+  NavigationTrackerProps
+>(({children, provider}, ref) => {
+  // Initializing a Trace instance
+  const tracer = useTrace({name: 'navigation', version: '1.0.0'}, provider);
```
jpmunz, line 19: you may be able to read the name + version from package.json actually

packages/experimental/navigation/hooks/__tests__/useNavigationTracker.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+    const mockStateListenerCall = mockAddListener.mock.calls[0][1];
+
+    mockGetCurrentRoute.mockReturnValue({name: 'first-view-test'});
+    mockStateListenerCall();
+    expect(mockSpanEnd).not.toHaveBeenCalled();
+    // starts the initial span
+    expect(mockStartSpan).toHaveBeenCalledTimes(1);
+
+    mockGetCurrentRoute.mockReturnValue({name: 'second-view-test'});
```
jpmunz, line 60: I'd be curious to take another look at these tests with less pieces mocked, e.g. I think we could use an actual tracerProvider and just need to supply it with a exporter that is mocked and in that way assert on the full body of the spans that would have been written out

packages/experimental/navigation/otel/hooks/useTrace.ts
facostaembrace: yes, I changed it for testing purposes but I just added `useLayoutEffect` back

packages/experimental/navigation/hooks/useNativeNavigationTracker.ts
facostaembrace: Nothing, you are right. I removed this block of code.

packages/experimental/navigation/components/NativeNavigationTracker.tsx
facostaembrace: Typescript complains if I do not add the fragment as wrapper here. It seems like the typescript version doesn't work well with the types from the current `react-native` version (it's a very old one, v0.56.0).

tsconfig.json
```diff
@@ -3,7 +3,7 @@
     "target": "es6",
     "module": "commonjs",
     "moduleResolution": "node",
-    "jsx": "react",
+    "jsx": "react-native",
```
facostaembrace, line 6: this is necessary for running tests that implement components

experimental/navigation/hooks/useNativeNavigationTracker.ts
```diff
@@ -0,0 +1,56 @@
...
+      spanCreator(tracer, span, navView, componentName);
+    });
+
+    navigationElRef.registerComponentDidDisappearListener(({componentName}) => {
+      if (!componentName) {
+        // do nothing in case for some reason there is no route
+        return;
+      }
+
+      spanEnd(span);
```
jpmunz, line 50: I forget if we already discussed this but we should think about how we want to handle the last view when the app ends or gets backgrounded. If we don't have a strong opinion maybe just adding a function to the hook's return value like `endLastSpan` or something and telling the user that if they care about ending it they should hook it up to a handler

package.json
```diff
@@ -53,6 +53,7 @@
     "react-native": ">=0.56.0"
   },
   "dependencies": {
+    "@opentelemetry/api": "^1.9.0",
```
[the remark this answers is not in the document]
facostaembrace: +1. I think I'll revisit it as a separated ticket (one into the KTLO list, as you mentioned). but totally agreed on it

experimental/navigation/types/navigation.ts
```diff
@@ -0,0 +1,18 @@
+interface TNavigationContainer {
```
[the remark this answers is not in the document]
facostaembrace: yeah, I saw in the past that people used `T` as prefix when describing a shape for typescript (but declaring that as a `type`). I think in here it would make sense to update the name in favor of the `I` prefix since this is declared using `interface` instead

experimental/navigation/hooks/useNativeNavigationTracker.ts
facostaembrace: we talked about it, but we didn't decide any approach. I like the suggestion, I'll dig into this on friday

PR #31, opened 2024-06-11
Update install script and uninstall script

TheMutsi:
## Goal

Update install scripts to support kotlin and swift, also update uninstall scripts


jpmunz (commented):
@TheMutsi I'm having a bit of trouble following the sequence of steps performed by the install scripts here, would it be possible to add a readme maybe in packages/core/scripts/setup that gives an overview of the steps here? Are there associated docs repo changes that describe what the updated manual steps would be? I could maybe take a look there as well

TheMutsi:
> @TheMutsi I'm having a bit of trouble following the sequence of steps performed by the install scripts here, would it be possible to add a readme maybe in packages/core/scripts/setup that gives an overview of the steps here? Are there associated docs repo changes that describe what the updated manual steps would be? I could maybe take a look there as well

Sure I will update with the entry endpoint script

TheMutsi:
node node_modules/@embrace-io/react-native/lib/scripts/setup/installIos.js
node node_modules/@embrace-io/react-native/lib/scripts/setup/installAndroid.js
Uninstall
node node_modules/@embrace-io/react-native/lib/scripts/postunlink/index.js

jpmunz:
> > @TheMutsi I'm having a bit of trouble following the sequence of steps performed by the install scripts here, would it be possible to add a readme maybe in packages/core/scripts/setup that gives an overview of the steps here? Are there associated docs repo changes that describe what the updated manual steps would be? I could maybe take a look there as well
> 
> Sure I will update with the entry endpoint script

Thanks for adding this! Taking another look

packages/core/scripts/setup/ios.ts
```diff
@@ -4,92 +4,78 @@ const path = require('path');
...
+    podfile.addBefore('use_react_native', `${embraceNativePod}\n`);
+
+    return podfile.patch();
+  });
+};
+
 export const iosPodfile = {
   name: 'Podfile patch (ONLY React Native Version < 0.6)',
   run: (wizard: Wizard): Promise<any> =>
-    wizard.fieldValue(packageJSON).then((json) => {
```
jpmunz, line 70: what was the reason for moving this above?

packages/core/scripts/setup/patches/common.ts
```diff
@@ -0,0 +1,85 @@
...
+  lineToAdd: string,
+  searchText: string | RegExp,
+  fileName?: string
+): boolean => {
+  if (file.hasLine(lineToAdd)) {
+    logger.warn(`${fileName} already contains the line: ${lineToAdd}`);
+    return false;
+  } else {
+    logger.warn(`Adding the line: ${lineToAdd} in ${fileName}`);
+    file.addBefore(searchText, lineToAdd);
```
jpmunz, line 62: might be simpler to have a single function and take 'before'|'after' as a parameter

packages/core/scripts/setup/patches/common.ts
```diff
@@ -0,0 +1,85 @@
+import EmbraceLogger from '../../../src/logger';
+import { FileUpdatable } from '../../util/file';
+import { OBJECTIVEC_APP_DELEGATE } from './ios/ios.objectivec';
+import { SWIFT_APP_DELEGATE } from './ios/ios.swift';
+
+// TODO refactor this
```
jpmunz, line 6: what change would be needed here?

packages/core/scripts/postunlink/README.md
```diff
@@ -0,0 +1,9 @@
+# Uninstall Script
+
+## Universal Uninstall Script
+
+```shell
+node node_modules/@embrace-io/react-native/lib/scripts/postunlink/index.js
+```
+
+This script call `postunlink.android` and `postunlink.ios` to revert all changes that we made in the install script.
```
jpmunz, line 9: could we use `uninstall` for the directory and all file names? Why was postunlink chosen?

packages/core/scripts/setup/patches/android/patch.java.ts
jpmunz, line 1: the naming of files in the patches folder is a bit inconsistent, e.g. 'ios.objectivec.ts' vs. 'patch.java.ts'

since we have camel case elsewhere maybe just `installJava.ts` + `uninstallJava.ts` for everything? Or even just `java.ts` `uninstallJava.ts`

jpmunz: overall it seems that there is potential to replace these files with a single function that operated on a structure that encoded these replacements, e.g. something like:

```
PATCH_OPERATIONS ... = {
    KOTLIN: {
      lines: [
        {find: 'some regex', add: 'after|before', text: 'text to add'},
         ...
      ]
    },
    SWIFT: ...
}
```

using the same structure for removing as well

packages/core/scripts/util/android.ts
```diff
@@ -95,7 +98,7 @@ export const mainApplicationPatchable = ({
 
     if (!mainApplicationPath) {
       return reject(
-        `cannot find MainApplication.java file in any folder at ${p}. Please refer to the docs at https://embrace.io/docs/react-native/integration/add-embrace-sdk/?rn-platform=android to update it manually.`
+        `cannot find ${MAIN_CLASS_BY_LANGUAGE[platform]} file in any folder at ${p}. Please refer to the docs at https://embrace.io/docs/react-native/integration/add-embrace-sdk/?rn-platform=android to update it manually.`
```
jpmunz, line 101: looks like the query string needs to be `?platform=android` to select the correct tab

packages/core/scripts/setup/ios.ts
TheMutsi: Unit test

packages/core/scripts/setup/ios.ts
```diff
@@ -4,92 +4,78 @@ const path = require('path');
...
 import EmbraceLogger from '../../src/logger';
 import {
-  appDelegatePatchable,
-  bundlePhaseExtraArgs,
   bundlePhaseRE,
-  embraceImport,
   embraceNativePod,
   embracePlistPatchable,
   embRunScript,
-  formatEmbraceInitializer,
```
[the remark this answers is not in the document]
TheMutsi: Yes! nice catch

packages/core/scripts/setup/patches/common.ts
TheMutsi: I wonder if we can use one a "static" approach. something like 
EmbraceLogger.warn|info|etc. without create an instance every file/every call

packages/core/scripts/setup/patches/common.ts
TheMutsi: Following single responsibility principle it should be 2 function. But I can change it if you want

packages/core/scripts/postunlink/README.md
TheMutsi: I can rename it and remove post unlink

packages/core/scripts/setup/android.ts
```diff
@@ -112,31 +112,20 @@ export const createEmbraceJSON = {
...
+  const response = patch('kotlin');
+  if (!response) {
+    return patch('java');
+  }
+  return response;
+};
+
 export const patchMainApplication = {
-  name: 'patch MainApplication.java file',
+  name: 'patch MainApplication.kt file',
```
jpmunz, line 125: could be either .java or .kt right? Maybe we just say "patch MainApplication file" with no extension

packages/core/scripts/setup/patches/patch.ts
```diff
@@ -0,0 +1,177 @@
...
+  ANDROID_LANGUAGE,
+  IOS_LANGUAGE,
+  MAIN_CLASS_BY_LANGUAGE,
+  SUPPORTED_LANGUAGES,
+} from './common';
+
+export const EMBRACE_IMPORT_SWIFT = 'import Embrace';
+
+export const EMBRACE_INIT_SWIFT =
+  'Embrace.sharedInstance().start(launchOptions: launchOptions, framework:.reactNative)';
```
jpmunz, line 22: I would put these strings directly in the `textToAdd` definitions below, it's a bit harder to see the patch that's being done when they're defined separately here

packages/core/scripts/setup/uninstall.ts
```diff
@@ -0,0 +1,337 @@
...
+
+interface IUnlinkEmbraceCode {
+  stepName: string;
+  fileName: string;
+  textsToDelete: ITextToDelete[];
+  findFileFunction: FindFileFunction;
+  docUrl: string;
+}
+
+const UNINSTALL_IOS_SWIFT_APPDELEGATE: IUnlinkEmbraceCode = {
```
jpmunz, line 61: these should share a common definition with what is in packages/core/scripts/setup/patches/patch.ts rather than being defined again since whatever text we're adding there should be removed here

packages/core/scripts/util/ios.ts
```diff
@@ -105,12 +108,13 @@ export const xcodePatchable = ({
     }
 
     if (!projectPath) {
+      console.log('SALIOOO');
```
jpmunz, line 111:
```suggestion
```

packages/core/scripts/setup/patches/patch.ts
TheMutsi: Do you mean direclty in the object ? 
textToAdd: `Embrace.sharedInstance().start(launchOptions: launchOptions, framework:.reactNative)\n`,

And copy this text and paste into the install/uninstall tests?

packages/core/scripts/__tests__/install.android.test.ts
```diff
@@ -25,7 +25,8 @@ describe('Modify Build Gradle', () => {
     let failed = 0;
     try {
       await wiz.processSteps();
-    } catch {
+    } catch (e) {
+      console.log('ERRR', e);
```
jpmunz, line 29:
```suggestion
```

packages/core/scripts/setup/patches/patch.ts
jpmunz: ya if they are only used in tests I think it's better to put it directly into `textToAdd` here and then just have the values literally written out in the tests

packages/core/scripts/setup/uninstall.ts
```diff
@@ -0,0 +1,305 @@
...
+
+interface IUnlinkEmbraceCode {
+  stepName: string;
+  fileName: string;
+  textsToDelete: ITextToDelete[];
+  findFileFunction: FindFileFunction;
+  docUrl: string;
+}
+
+const UNINSTALL_ANDROID_SWAZZLER_IMPORT: IUnlinkEmbraceCode = {
```
jpmunz, line 41: not a blocker but I think we should revisit this at some point as there's still opportunity for this to become out of sync with the install steps, I think ideally there would be just be a single list of operations each of which knowing how to do+undo itself

packages/core/scripts/util/file.ts
```diff
@@ -49,6 +53,53 @@ class FileContents implements FileUpdatable {
     this.contents = fs.readFileSync(path, 'utf8');
   }
 
+  public getPaddingFromString(searchString: string): string {
```
jpmunz, line 56: could both of these be handle by a regex? Something like:
```
const paddingBefore = /\n(\s*)/
const paddingAfter = /(\s*)\n/
const search = new Regexp(paddingBefore.source + searchString + paddingAfter.search, 'm');
```

and then the matching groups for the search would contain both the before and after padding?

jpmunz (approved):

PR #32, opened 2024-06-11
Changes in spans

TheMutsi:
## Goal

Add startTime and endTime to spans

### DIFF
--- a/packages/core/android/src/main/java/io/embrace/embracewrapper/EmbraceManagerModule.java
+++ b/packages/core/android/src/main/java/io/embrace/embracewrapper/EmbraceManagerModule.java
@@ -569,9 +569,13 @@ public class EmbraceManagerModule extends ReactContextBaseJavaModule {
     }
 
     @ReactMethod()
-    public void startSpan(String name, String parentSpanId, Promise promise) {
+    public void startSpan(String name, String parentSpanId, Double startTimeMS, Promise promise) {
         try{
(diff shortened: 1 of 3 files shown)


packages/core/android/src/main/java/io/embrace/embracewrapper/EmbraceManagerModule.java
```diff
@@ -569,9 +569,13 @@ public void logNetworkClientError(String url,
     }
 
     @ReactMethod()
-    public void startSpan(String name, String parentSpanId, Promise promise) {
+    public void startSpan(String name, String parentSpanId, Double startTimeMS, Promise promise) {
```
jpmunz, line 572: technically this is nanoseconds correct? Should the parameter be startTimeNano and endTimeNano below?

TheMutsi: I left this as MS since android already accept MS I am waiting for iOS to accept ms and remove the MS=>Nano conversion. 
Do you want to change it meanwhile ?

jpmunz: hmm but even if Android is accepting milliseconds this attribute is still nano seconds because `convertMSToNano` is done on the JS regardless of platform right? I think for clarity I would either rename this to Nano for now or on the JS side do a condition based on platform and only call `convertMSToNano` for ios

TheMutsi: sure done

PR #34, opened 2024-06-13
[EMBR-3984] Improve linting in repo

facostaembrace:
## Goal

- removed `tslint`
- added `eslint` and `prettier`.

## Note
We have to re-visit the rules that I added and turned "off" by default since those were throwing errors and the goal here wasn't to refactor any code.


tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
-    "target": "es6",
-    "module": "commonjs",
+    "target": "es2021",
```
jpmunz, line 3: this seems like a bigger change, was it related to the linting?

jpmunz (commented):
awesome!

tsconfig.json
facostaembrace: actually I wanted to upgrade also these configs but at the end of the day I realized that was breaking a test, so reverted the change to how it was before

jpmunz (approved):

PR #36, opened 2024-06-17
EMB-3907 Android side for tracer provider

jpmunz:
Adding a react-native-tracer-provider package. This provides a wrapper on the `TracerProvider` interface exposed by the (android) native SDK that also conforms to the `TracerProvider` interface on the JS side so that any JS otel instrumentation library should be able to use the Embrace SDK to write traces.

Keeping in draft for now as there is a number of TODOs/blockers atm:
* Need to implement the iOS side of the native module, this is waiting on pointing to the latest 6.x version of the iOS Embrace SDK
* Need to point to an updated version of the Android Embrace SDK that exposes the TracerProvider interface, this is under development at the moment, in the meantime I used the OTEL Java SDK for testing


packages/react-native-tracer-provider/android/src/main/java/io/embrace/tracerprovider/TracerProviderModule.kt
```diff
@@ -0,0 +1,288 @@
...
+        val map = this.writableMapBuilder.build()
+
+        map.putString(SPAN_CONTEXT_TRACE_ID_KEY, spanContext?.traceId ?: "")
+        map.putString(SPAN_CONTEXT_SPAN_ID_KEY, spanContext?.spanId ?: "")
+
+        return map
+    }
+
+    init {
+        // TODO replace with Embrace OTEL provider
```
alyssa-embrace, line 144: Is this comment up to date?

packages/react-native-tracer-provider/android/src/main/java/io/embrace/tracerprovider/TracerProviderModule.kt
```diff
@@ -0,0 +1,288 @@
...
+
+        tracers[id] = tracerProvider.get(name, version)
+    }
+
+    @ReactMethod fun startSpan(tracerName: String, tracerVersion: String, spanBridgeID: String,
+                               name: String, kind: String, time: Double,
+                               attributes: ReadableMap, links: ReadableArray, parentID: String,
+                               promise: Promise) {
+        val tracer = tracers["$tracerName $tracerVersion"] ?: return;
+        val spanBuilder = tracer.spanBuilder(name);
```
alyssa-embrace, line 182: Is this semi-colon necessary? It doesn't seem to match the style of everything else around it.

packages/react-native-tracer-provider/android/src/main/java/io/embrace/tracerprovider/TracerProviderModule.kt
```diff
@@ -0,0 +1,288 @@
...
+        val statusCode = status.getString(SPAN_STATUS_CODE_KEY) ?: return
+        val message = status.getString(SPAN_STATUS_MESSAGE_KEY) ?: ""
+
+        try {
+            if (message.isEmpty()) {
+                span.setStatus(StatusCode.valueOf(statusCode))
+            } else {
+                span.setStatus(StatusCode.valueOf(statusCode), message)
+            }
+        } catch(e: IllegalArgumentException) {}
```
alyssa-embrace, line 271: Can we log this rather than just swallowing it?

packages/react-native-tracer-provider/src/EmbraceNativeSpan.ts
```diff
@@ -0,0 +1,224 @@
...
+  isAttributes,
+  normalizeAttributes,
+  normalizeLinks,
+  normalizeTime,
+} from './util';
+
+/**
+ * EmbraceNativeSpan implements a Span over the native Embrace Android and iOS SDKs.
+ *
+ * A handful of simple attributes are maintained on the JS side for each span include a unique ID based on the
```
alyssa-embrace, line 24: I feel like this should be `including` instead of `include`

packages/react-native-tracer-provider/src/EmbraceNativeSpan.ts
```diff
@@ -0,0 +1,224 @@
...
+        normalizeAttributes({}),
+        normalizeTime(attributesOrStartTime)
+      );
+    }
+
+    return this;
+  }
+
+  public addLink(link: Link): this {
+    return this.addLinks([link]);
```
alyssa-embrace, line 143: Could we save an allocation on boxing the link in an array by inverting `addLinks` and `addLink`? That is, by writing this such that `addLinks` calls `addLink` rather than this way around?

packages/react-native-tracer-provider/src/EmbraceNativeSpan.ts
```diff
@@ -0,0 +1,224 @@
...
+
+    TracerProviderModule.endSpan(this.nativeID(), normalizeTime(endTime));
+    this.recording = false;
+  }
+
+  public isRecording(): boolean {
+    return this.recording;
+  }
+
+  public recordException(exception: Exception, time?: TimeInput): void {
```
alyssa-embrace, line 192: Is there a reason this one doesn't return `this` like some of the ones above it?

integration-tests/basic-test-app/helpers/generateSpans.ts
```diff
@@ -0,0 +1,83 @@
...
+/* eslint-disable @typescript-eslint/no-unused-vars */
+import {
+  context,
+  SpanKind,
+  SpanStatusCode,
+  trace,
+  Tracer,
+} from "@opentelemetry/api";
+
+export function generateTestSpans(tracer: Tracer) {
```
alyssa-embrace, line 12: Random thought: is it possible to construct a span where the end-time is before the start-time?

packages/react-native-tracer-provider/src/util.ts
```diff
@@ -0,0 +1,55 @@
...
+export const normalizeTime = (time?: TimeInput): number => {
+  if (time === undefined) {
+    return 0;
+  }
+
+  if (typeof time === 'number') {
+    return time;
+  }
+
+  if (time instanceof Date) {
```
alyssa-embrace, line 19: Why are type checking semantics on 19 and 15 different?

integration-tests/basic-test-app/helpers/generateSpans.ts
```diff
@@ -0,0 +1,83 @@
...
+  if (span1.isRecording()) {
+    span1.end();
+  }
+  if (span1.isRecording()) {
+    span1.setAttribute("after-recording", "should not exist");
+  }
+
+  // Add some span events
+  span2.addEvent("test-2-event-1");
+  span2.addEvent("test-2-event-2", 1700001002000);
```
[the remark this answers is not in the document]
jpmunz: nope, just arbitrary values that can we can assert got added later in the test

integration-tests/basic-test-app/helpers/generateSpans.ts
```diff
@@ -0,0 +1,83 @@
...
+  SpanStatusCode,
+  trace,
+  Tracer,
+} from "@opentelemetry/api";
+
+export function generateTestSpans(tracer: Tracer) {
+  const span1 = tracer.startSpan("test-1");
+  const span2 = tracer.startSpan("test-2", { kind: SpanKind.CLIENT });
+  const span3 = tracer.startSpan("test-3");
+  // Not ended so shouldn't be part of the output
```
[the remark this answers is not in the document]
jpmunz: they should be independent the idea would be to call generateTestSpans followed by ending the session so the payload can be verified then separately in a following test we'll call generateNestedSpans and end that session so the payload shouldn't include data from the previous run

integration-tests/basic-test-app/helpers/generateSpans.ts
jpmunz: that's a good point I can add an example of that to see how it's handled, I believe the interface allows that so it would be up to our implementation how it would treat that case

integration-tests/basic-test-app/helpers/generateSpans.ts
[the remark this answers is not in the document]
alyssa-embrace: Ah OK. That makes sense. Is that documented or implicit anywhere that I missed?

integration-tests/specs/tracer_provider.test.ts
```diff
@@ -0,0 +1,181 @@
...
+      expectValidIDs(spans[2]);
+
+      expect(sessionPayloads[0].spans).toEqual([
+        {
+          trace_id: spans[0].trace_id,
+          span_id: spans[0].span_id,
+          parent_span_id: "0000000000000000",
+          name: "test-1-updated",
+          start_time_unix_nano: 1718045981827000 * 1000000,
+          end_time_unix_nano: 1718045981827000 * 1000000 + 2591701,
```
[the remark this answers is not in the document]
jpmunz: I haven't verified what the exact values should be yet so they might need to be tweaked but the idea would be to add the duration (e.g. 2591701) to the start time in order to get what the end time should be, I can add a comment

integration-tests/typings/embrace.d.ts
```diff
@@ -2,16 +2,41 @@
 // add more properties as they become relevant for test specs, see:
 // https://github.com/embrace-io/embrace-android-sdk/tree/master/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/payload
 
-interface SessionPayload {
+interface Session {
```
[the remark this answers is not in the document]
jpmunz: ya I was moving stuff around a bit to align with the naming from the android SDK, I may rename `SessionMessage` -> `SessionPayload` actually but that represents the full payload sent to the session endpoint, inside that payload there is a key ("s") that represents fields about the session itself

integration-tests/specs/tracer_provider.test.ts
```diff
@@ -0,0 +1,181 @@
+import { driver } from "@wdio/globals";
+
+import { getSessionMessages } from "../helpers/embrace_server";
+import { EmbraceSpanData } from "../typings/embrace";
+
+describe("Tracer Provider", () => {
+  const expectValidIDs = (span: EmbraceSpanData) => {
+    expect(span.span_id).toHaveLength(16);
+    expect(span.span_id).not.toEqual("0000000000000000");
```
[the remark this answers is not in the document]
jpmunz: ya this is the body of `expectValidIDs`, the actual format is defined in the otel spec: https://opentelemetry.io/docs/specs/otel/overview/#spancontext

packages/react-native-tracer-provider/android/src/main/java/io/embrace/tracerprovider/TracerProviderModule.kt
jpmunz: yep that's part of why this is still in draft this is still using just the OTEL SDK for prototyping purposes it will need to be updated to use the actual Embrace one

packages/react-native-tracer-provider/android/src/main/java/io/embrace/tracerprovider/TracerProviderModule.kt
jpmunz: nope! There's a few that I see now, just my habit from the JS side I'll remove

packages/react-native-tracer-provider/android/src/main/java/io/embrace/tracerprovider/TracerProviderModule.kt
jpmunz: will add

packages/react-native-tracer-provider/src/EmbraceNativeSpan.ts
jpmunz: yep

packages/react-native-tracer-provider/src/EmbraceNativeSpan.ts
jpmunz: this is driven by the OTEL JS API so it's forced to follow this interface: https://github.com/open-telemetry/opentelemetry-js/blob/main/api/src/trace/span.ts#L149

packages/react-native-tracer-provider/src/util.ts
jpmunz: `typeof` only works for primitive types since Date is a class typeof would just return "object" without giving more info so need `instanceof` to see that it is actually a Date

integration-tests/basic-test-app/helpers/generateSpans.ts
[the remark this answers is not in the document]
jpmunz: let me add some comments to this file, I wasn't sure if I was going to stick with this approach as there were 2 options:
1. in the basic test app expose buttons/inputs for interacting with the spans API so that the test case can do all the individual setup (e.g. start span, add attributes, etc.)
2. in the basic test app expose a single "test this scenario" button and do the asserts in the test case

1) I think is ideal because then you can read all the setup and assertions together in the test, however when I started to implement this way I found it was just very time consuming to create the full interface that was needed to do all the setup that is being done here

2) simplifies things but creates this disconnect between the assertions in the test spec and what setup was actually done, probably going to stick with this for now but I'll add comments to this file to make it clear why that was separated

packages/react-native-tracer-provider/src/EmbraceNativeSpan.ts
alyssa-embrace: Makes sense. Thanks for clarifying!

integration-tests/specs/tracer_provider.test.ts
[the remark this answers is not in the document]
alyssa-embrace: I am a dork. Ignore me.

packages/react-native-tracer-provider/android/src/main/java/io/embrace/tracerprovider/TracerProviderModule.kt
```diff
@@ -0,0 +1,288 @@
...
+     */
+
+    private val LINK_ATTRIBUTES_KEY = "attributes";
+    private val LINK_SPAN_CONTEXT_KEY = "context";
+    private val SPAN_CONTEXT_TRACE_ID_KEY = "traceId";
+    private val SPAN_CONTEXT_SPAN_ID_KEY = "spanId";
+    private val SPAN_STATUS_CODE_KEY = "code";
+    private val SPAN_STATUS_MESSAGE_KEY = "message";
+
+    private fun stringListFromReadableArray(array: ReadableArray) :List<String> {
```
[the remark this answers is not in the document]
alyssa-embrace: Oh OK. So it fits existing semantics.

packages/react-native-tracer-provider/android/src/main/java/io/embrace/tracerprovider/TracerProviderModule.kt
```diff
@@ -0,0 +1,288 @@
...
+                ReadableType.Number -> builder.put(key, map.getDouble(key))
+                ReadableType.String -> builder.put(key, map.getString(key) ?: "")
+                ReadableType.Array -> {
+                    val array = map.getArray(key)
+                    val arrayType = array?.getType(0);
+                    when (arrayType) {
+                        ReadableType.Boolean -> builder.put(AttributeKey.booleanArrayKey(key), booleanListFromReadableArray(array))
+                        ReadableType.Number -> builder.put(AttributeKey.doubleArrayKey(key), doubleListFromReadableArray(array))
+                        ReadableType.String -> builder.put(AttributeKey.stringArrayKey(key), stringListFromReadableArray(array))
+                        else -> continue
```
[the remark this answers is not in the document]
alyssa-embrace: Makes sense then. We can capture it later perhaps.

alyssa-embrace (approved):
All my questions have been addressed.

jpmunz:
going to close this and re-open, already split up the changes a bit and want to further separate the changes between the JS and Android side

PR #37, opened 2024-06-19
[EMBR-3873] Instrumentation library for Redux

facostaembrace:
## Goal

Adding a method intended to be a middleware that will be when Redux is in use to create telemetry data about actions.

Pending:
- improve tests (kind of placeholder now).
- isolate dependencies (should move `redux` dev dependency from the root package.json into the experimental package)
- discuss about what we want to log into the span when an action is dispatched (do we want to add attributes? events? what else?)

NOTE: will point to `master` as soon as the [Navigation PR](https://github.com/embrace-io/embrace-react-native-sdk/pull/28) is merged.

## Testing

TBD more assertions for unit tests


experimental/navigation/otel/spanCreator.ts
```diff
@@ -10,16 +10,16 @@ const ATTRIBUTES = {
 const spanStart = (
   tracer: TracerRef,
   span: SpanRef,
-  currentRouteName: string,
-  isFirstView: boolean,
+  name: string,
+  isFirstView?: boolean,
```
jpmunz, line 14: if we're re-using this probably `isFirstView` could be moved out as well and have the caller set the attributes. Another option for keeping this independent would be to duplicate the helpers from navigation into here and then refactor them later if we have a standalone helper package

experimental/redux/otelMiddleware.ts
```diff
@@ -0,0 +1,42 @@
...
+  name: string;
+  version: string;
+}
+
+const otelMiddleware = <RootState>(
+  provider: TracerProvider,
+  config: ConfigArgs,
+): Middleware<object, RootState> => {
+  if (!provider) {
+    console.warn("Provider is not available. Spans won't be recorded");
```
jpmunz, line 20: similar to my comment on the navigation package I think we should be fine always grabbing the global tracer if there's no provider specified, if nothing is setup then the span stuff will all just be no-ops

experimental/redux/otelMiddleware.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  const tracer = trace.getTracer(config.name, config.version);
+
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-expect-error
+  return (_: MiddlewareAPI<Dispatch<UnknownAction>, RootState>) => {
+    return (next: Dispatch<UnknownAction>) => {
+      return (action: Action) => {
+        const span = tracer.startSpan("redux-action");
```
jpmunz, line 30: taking a look at how middleware works we should maybe revisit the instrumentation here, it's a bit unfortunate that it works as a chain so it seems you really only get insight into how long the "next" middleware took but not how long the whole action took to work through the dispatch operation

there's maybe some potential alternatives with either a store enhancer or subscribing to change events:
https://redux.js.org/understanding/thinking-in-redux/glossary#store-enhancer
https://redux.js.org/api/store#subscribelistener
https://github.com/reduxjs/redux/issues/303#issuecomment-125184409

or if we kept to this approach maybe makes more sense as a log instead of span to record the single event

PR #40, opened 2024-06-25
EMBR-4049 Cleanup workspaces

jpmunz:
* upgrade to Yarn Modern (https://yarnpkg.com/migration/overview)
* use Yarn Constraints to enforce that all packages have consistent metadata
* remove unused scripts commands
* remove docs


package.json
```diff
@@ -1,32 +1,18 @@
...
+  "private": true,
+  "description": "Monorepo for packages that power Embrace's React Native SDK.",
+  "workspaces": [
+    "packages/*"
   ],
+  "packageManager": "yarn@4.3.1",
+  "engines": {
+    "node": ">=18.0.0"
+  },
   "scripts": {
```
jpmunz, line 12: @TheMutsi I was hoping to get rid of old scripts here, are there any I removed that are still being used? Some additional questions:
- I removed the `-tag-version-prefix='rn-os-modules-'` part of the publish command, is it needed?
- should "build" be changed to "npx lerna run tsc" or "npx lerna run build"? I see that it generated a `/dist` folder at the root level but I assumed we were publishing the ones in packages/*/dist. The CI workflow also does a `yarn build` after running lint and test that I wasn't sure if it was needed
- I moved 'prettier' to be part of 'lint' as it didn't seem like that was being run as part of the CI previously

package.json
```diff
@@ -37,29 +23,9 @@
...
-    "react-native",
-    "react-component",
-    "tracking",
-    "embrace",
-    "bug tracker"
-  ],
-  "peerDependencies": {
-    "react-native": ">=0.56.0"
-  },
-  "dependencies": {
```
jpmunz, line 55: moved to packages/core

packages/core/package.json
```diff
@@ -29,48 +10,49 @@
...
+  },
+  "main": "lib/index.js",
+  "typings": "lib/index.d.ts",
+  "directories": {
+    "lib": "lib",
+    "test": "__tests__"
+  },
+  "rnpm": {
+    "commands": {
+      "postlink": "node node_modules/embrace-io/dist/scripts/postlink",
```
jpmunz, line 42: @TheMutsi were these postlink/unlink commands still needed? I wasn't sure if the path was correct if they were part of the core package I thought that would belong under node_moudles/@embrace-io/react-native/...

packages/core/package.json
```diff
@@ -29,48 +10,49 @@
...
+      "postunlink": "node node_modules/embrace-io/dist/scripts/postunlink"
+    }
   },
-  "gitHead": "0e15bb316225a075d6dc2437b9784b441d6c0533"
```
jpmunz, line 75: @TheMutsi I removed `gitHead` from each package.json, was it being used?

react-native.config.js
jpmunz, line 1: @TheMutsi was this needed for anything?

jpmunz (commented):
Couple additional questions:
* is the `Pipfile` in the project root needed?
* for all the packages other than core would it make sense to have them define a peerDependency on @embrace-io/react-native?

package.json
TheMutsi: `-tag-version-prefix='rn-os-modules-` was needed because it was giving errors with tags when we deploy the NPM core and modules.
You can remove yarn build, the last change that I made in the repo was moving the principal SDK to a lerna package called core. Now I see there are some scripts that I forgot to delete. The most important is publish-modules now, then test/jest, linters etc

packages/core/package.json
TheMutsi: Yep it was used to specify a specific commit when you publish a new NPM version.

packages/core/scripts/__tests__/__mocks__/buildWithoutSwazzler.gradle
```diff
@@ -38,6 +38,12 @@ buildscript {
         
         
         
+        
+        
+        
+        
+        
+
```
TheMutsi, line 46: there are some empty breaking lines, idk if this was intentional

packages/spans/package.json
```diff
@@ -2,40 +2,40 @@
...
+  "author": "Embrace <support@embrace.io> (https://embrace.io/)",
+  "bugs": {
+    "url": "https://github.com/embrace-io/embrace-react-native-sdk/issues",
+    "email": "support@embrace.io"
   },
-  "homepage": "https://embrace.io/",
-  "license": "ISC",
-  "main": "lib/src/index.js",
-  "typings": "lib/src/index.d.ts",
+  "homepage": "https://github.com/embrace-io/embrace-react-native-sdk",
```
TheMutsi, line 10: I think "react-native": ">=0.56.0" is missing for spans

yarn.config.cjs
```diff
@@ -0,0 +1,121 @@
...
+    } else {
+      workspace.set("keywords", [...new Set([
+        ...(workspace.keywords || []),
+        "embrace",
+        "react-native",
+        "tracking",
+        "observability",
+        "instrumentation",
+        "telemetry",
+        "bug tracker",
```
TheMutsi, line 51: Now that it is a yarn workspace you can se this, dependencies and peerDependency, hash and other things from here instead of having this in each package json.
You can delete this + peer dependencies + other common things from the inidivual json packages

TheMutsi (changes requested):

packages/core/package.json
jpmunz: is this added automatically by some part of the release process?

facostaembrace:
@TheMutsi do you have any suggestion about how to test these changes? We would like to know if there is a way of publishing a release but in an "staging" environment (instead of "production" directly). Do you know if that would be possible into the npm registry? Our goal would be to release and test but in a "private" mode, we don't want users to upgrade or make the new version to show up until we are confident about the changes (maybe a beta version?)

package.json
jpmunz: got it, okay I removed the rest and just kept `publish`, `lint, `test`. I removed `tag-version` for now and we can see what errors we get when trying to deploy

yarn.config.cjs
jpmunz: this works as a linter/validation steps so the actual values do end up getting written to the individual `package.json` files but that should happen automatically we'll just have to worry about keeping this file up to date

packages/spans/package.json
jpmunz: added this to the yarn constraints so now each package.json should have it, is it fine that everything in packages/* specifies it now?

packages/core/package.json
TheMutsi: No, we can delete it

jpmunz: deleted

TheMutsi:
@facostaembrace I have never tried but you can publish an NPM module as private and then grant R/W access to some specific users. Also you can create beta/alpha/etc tags and publish it. Only the `latest`(this is the name - latest) tag which is the default one will be auto-updated on install

packages/spans/package.json
TheMutsi: nono its fine

react-native.config.js
TheMutsi: It is created by default in SDK, to specify some configuration.

packages/core/package.json
TheMutsi: It should be added automatically after publish a new version with the commit hash of the version submitted

jpmunz:
> @facostaembrace I have never tried but you can publish an NPM module as private and then grant R/W access to some specific users. Also you can create beta/alpha/etc tags and publish it. Only the `latest`(this is the name - latest) tag which is the default one will be auto-updated on install

I added this to https://www.notion.so/embraceio/Integrate-releases-and-changelog-updates-with-CI-CD-ab93efad962d4e6aba692dfb5d9dba81?pvs=4, I think could be interesting as a "staging" workflow to be able to do this

react-native.config.js
jpmunz: that makes sense I'll add it back as it was

facostaembrace (approved):

yarn.config.cjs
```diff
@@ -0,0 +1,126 @@
...
+/**
+ *  Enforces each package having a peerDependency on React Native
+ */
+function enforceReactNativePeerDependency({ Yarn }) {
+  for (const workspace of Yarn.workspaces()) {
+    if (workspace.manifest.private) continue;
+
+    workspace.set("peerDependencies", {
+      ...(workspace.peerDependencies || {}),
+      "react-native": ">=0.56.0",
```
TheMutsi, line 82: Since you added this here, you could remove this peer dependency from the individual pkg

yarn.config.cjs
```diff
@@ -0,0 +1,126 @@
...
+
+/**
+ *  Enforces each package having a common set of devDependencies
+ */
+function enforceCommonDevDependencies({ Yarn }) {
+  for (const workspace of Yarn.workspaces()) {
+    if (workspace.manifest.private) continue;
+
+    workspace.set("devDependencies", {
+      "typescript": "^4.7.4",
```
TheMutsi, line 95: Same here

.github/workflows/node.js.yml
```diff
@@ -21,6 +21,7 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
           cache: "npm"
+      - run: corepack enable
```
TheMutsi, line 24: Just curious, why did you add this ?

TheMutsi (approved):
lgtm

.github/workflows/node.js.yml
jpmunz: this is a required setup step for the latest yarn: https://yarnpkg.com/getting-started/install

yarn.config.cjs
jpmunz: the value still ends up written to the individual package.json files, this just adds it as a linting step to make sure they all have the same and can auto fix it if it sees a discrepancy (when `yarn constraints --fix` gets run)

PR #42, opened 2024-06-26
Fix extra lines added to test mock

jpmunz:
The test was adding and removing a line to the mock file but in removing it it was leaving behind trailing whitespace which ended up as part of git changes

### DIFF
--- a/packages/core/scripts/__tests__/scripts.test.ts
+++ b/packages/core/scripts/__tests__/scripts.test.ts
@@ -6,8 +6,10 @@ beforeEach(() => {
   jest.clearAllMocks().resetModules();
 });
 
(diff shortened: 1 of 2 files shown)


packages/core/scripts/__tests__/scripts.test.ts
```diff
@@ -39,25 +45,7 @@ describe("Modify Build Gradle", () => {
     file.patch();
     expect(file.hasLine(androidEmbraceSwazzler)).toBe(false);
   });
-  test("Update Android Swazzler Version", async () => {
```
jpmunz, line 42: this test seemed to be a repeat of the previous one

facostaembrace (approved):

PR #43, opened 2024-07-04
Rn 4.2.0

TheMutsi:
## Goal

Add Start time to start span
Add End Time to stop span
Add install script for swift and kotlin
Add required startTimeMS for addSpanEventToSpan
Fix uninstall script
Fix navigation ref type for react-navigation
Fix Lint adding lib to ignore folders because they are autogenerated files and it was breaking the pre commit hooks and the publish modules
Bump Lerna to V8 and fix Lerna config for yarn workspace


packages/spans/src/index.ts
```diff
@@ -47,10 +57,10 @@ export const stopSpan = (
 export const addSpanEventToSpan = (
   spanId: string,
   name: string,
-  timeStampMs?: number,
```
jpmunz, line 50: shouldn't this still be optional?

jpmunz (approved):

packages/spans/src/index.ts
TheMutsi: Michiel asked to mark this mandatory fo both, since iOS won't work on it in 5.x

PR #44, opened 2024-07-05
EMBR-3908 ios tracer provider

jpmunz:
Builds on `jpmunz/EMBR-3907-add-trace-provider-export` and adds the iOS side of the native implementation, similar to the Android side this is using the OTEL Swift SDK as a placeholder to test out the implementation, will need to swap this for the Embrace SDK when it's ready

A couple things to figure out:
* OTEL Swift API has an error around supporting bool array attributes 
* no way to set schema URL in the OTEL Swift API? Should we restrict this on the JS side for iOS?
* no way to add links to an existing span in the OTEL Swift API? Should we restrict this on the JS side for iOS?


jpmunz:
a bit too many conflicts with the changes I've made to the feature branch, will re-open with a cleaner state

PR #53, opened 2024-07-12
Feature react native tracer provider

jpmunz:
Feature branch for the react-native-tracer-provider package

Before merging:
* make sure the package versions in this PR match `main`
* add to packages list on README.md

After merging:
* create a release branch and release a 5.x version of this tracer provider package


packages/react-native-tracer-provider/android/gradle.properties
```diff
@@ -0,0 +1,5 @@
+TracerProvider_kotlinVersion=1.7.0
```
facostaembrace, line 1: is there a reason why we need a prefix here? `TracerProvider_`

packages/react-native-tracer-provider/native-src/android/app/dependencies.gradle
```diff
@@ -0,0 +1,4 @@
+dependencies {
+  api "io.embrace:embrace-android-sdk:6.13.0"
```
facostaembrace, line 2: do we want to update this version manually here in the future?

packages/react-native-tracer-provider/native-src/android/app/src/main/res/mipmap-hdpi/ic_launcher.png
facostaembrace, line 1: maybe we can cleanup all these files, I think everything into this `mipmap-hdpi` was auto generated, right?

packages/react-native-tracer-provider/react-native-tracer-provider.podspec
facostaembrace, line 1: in my `react-native-otlp` package I ended up creating a file called RNEmbraceOTLP.podspec. maybe we can agree on a naming convention for this type of files

packages/react-native-tracer-provider/src/EmbraceNativeTracerProvider.ts
```diff
@@ -0,0 +1,143 @@
...
+ * useEmbraceNativeTracerProvider makes sure that the Embrace SDK has been installed and started and
+ * then sets up a EmbraceNativeTracerProvider
+ *
+ * The EmbraceNativeTracerProvider implements a TracerProvider over the native Embrace Android and iOS SDKs.
+ * Thin wrapped objects representing Tracers and Spans are maintained at the JS level and use Native Modules to
+ * call down to the SDKs to perform the actual operations on them.
+ *
+ * The JS side of its implementation is modelled after [opentelemetry-sdk-trace-base](https://github.com/open-telemetry/opentelemetry-js/tree/main/packages/opentelemetry-sdk-trace-base)
+ */
+export const useEmbraceNativeTracerProvider = (
```
facostaembrace, line 30: can we move the hook into a different place? as per the name of this file (`EmbraceNativeTracerProvider`) I though I was going to see only a class but it is also containing the hook. I would suggest to move the hook to its own place `useEmbraceNativeTracerProvider`

packages/react-native-tracer-provider/src/types.ts
```diff
@@ -0,0 +1,22 @@
+import {TracerProvider} from "@opentelemetry/api";
+
+export interface UseEmbraceNativeTracerProviderResult {
```
facostaembrace, line 3: can we remove the `Use` prefix from the name for this interface?

packages/core/update_android_version.ts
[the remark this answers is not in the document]
jpmunz: makes sense, I filed https://www.notion.so/embraceio/Move-helper-scripts-to-common-scripts-folder-1207e3c9985280c39e64d3bf66f4bbdd?pvs=4 to move this and some others to scripts/

packages/react-native-tracer-provider/native-src/android/app/dependencies.gradle
jpmunz: this ends up getting rewritten by the `update_android_version.ts` script

packages/react-native-tracer-provider/android/gradle.properties
jpmunz: I had copied and pasted this from another example, I believe the reasoning is to distinguish it from another "kotlinVersion" property that may have been defined by the app that uses this package, e.g. in `packages/react-native-tracer-provider/android/build.gradle` we have:

```
def kotlin_version = rootProject.ext.has("kotlinVersion") ? rootProject.ext.get("kotlinVersion") : project.properties["TracerProvider_kotlinVersion"]
```

So we use a bare "kotlinVersion" property if it's available and then only fallback to this "TracerProvider_kotlinVersion" if it's not

packages/react-native-tracer-provider/src/types.ts
jpmunz: I wasn't sure about the convention here, this is the type of the return value of `useEmbraceNativeTracerProvider` so I was naming it to match, what do you think?

packages/react-native-tracer-provider/react-native-tracer-provider.podspec
jpmunz: ya let's discuss this one a bit more, I when with the `react-native-tracer-provider.podspec` to better match the package name though since we have `RNEmbrace.podspec` in core it is a bit inconsistent

packages/react-native-tracer-provider/src/EmbraceNativeTracerProvider.ts
jpmunz: makes sense, moved to its own `useEmbraceNativeTracerProvider.ts` file

packages/react-native-tracer-provider/native-src/android/app/src/main/res/mipmap-hdpi/ic_launcher.png
jpmunz: looks like most aren't required, deleted

packages/react-native-tracer-provider/src/types.ts
TheMutsi: I think the convention is remove the use and just let EmbraceNativeTracerProvider (A least I saw this way in other sdks)

package.json
```diff
@@ -17,8 +17,9 @@
     "lint:clang": "./scripts/run-clang-format.sh",
     "lint:swift": "./scripts/run-swiftlint.sh",
     "test": "jest",
+    "android:test": "npx lerna run android:test",
     "ios:install": "npx lerna run ios:install",
-    "ios:test": "npx lerna run ios:test"
+    "ios:test": "npx lerna run ios:test --concurrency=1"
```
TheMutsi, line 22: What does this concurrency ? Prevent run more than 1 test at the same time ?

jpmunz: it prevents more than one package running `ios:test` at the same time, I was finding when core + this one ran their ios tests together that there were failures I guess due to the simulator on the github runner trying to do two things at once and/or too many resources being used

TheMutsi: I see, I think this does the same that maxInstances: 1 in capabilities inside the wdio config. 
```

{
      platformName: "iOS",
      "appium:automationName": "XCUITest",
      "appium:deviceName": "iPhone 16 Pro",
      "appium:bundleId": "io.embrace.basictestapp",
      "appium:noReset": true,
       maxInstances: 1,
     }
```

jpmunz: ya similar idea, this is running the swift unit tests though not the integration tests through web driver those are still not part of the GHA workflows

TheMutsi: oh I see

packages/react-native-tracer-provider/src/types.ts
facostaembrace: I usually name this kind of instances as `EmbraceNativeTracerProviderReturn` (or `EmbraceNativeTracerProviderResult` as you have here but without the prefix)

jpmunz: updated

packages/react-native-tracer-provider/react-native-tracer-provider.podspec
jpmunz: updated to ReactNativeTracerProvider.podspec, add some notes to our consistency task to look at things like RN vs. ReactNative and including "Embrace" in the name later on

facostaembrace (approved):

PR #54, opened 2024-07-12
EMBR-3908 integrate android tracer provider

jpmunz:
Implementation of the Android native module for the RN tracer provider

@embrace-io/sdk-android I tried to keep this PR to just the Android changes to make it easier to review but here's some extra context if required:
* The JS side of this package exposes a Tracer Provider that conforms to the OTEL interface but under the hood it just keeps track of reference IDs for the tracers/spans and uses those to communicate with this module asking the native tracer provider to do the actual work of creating those objects
* PR that updates the integration tests for this after testing with the Android SDK: https://github.com/embrace-io/embrace-react-native-sdk/pull/50
  * test setup [here](https://github.com/embrace-io/embrace-react-native-sdk/blob/4ac0bfc9a79bd6721708e897696de99b081fd509/integration-tests/basic-test-app/helpers/generateSpans.ts) and assertions on payload [here](https://github.com/embrace-io/embrace-react-native-sdk/blob/4ac0bfc9a79bd6721708e897696de99b081fd509/integration-tests/specs/tracer_provider.test.ts)
* PR that updates the JS side of the provider after testing with the Android SDK: https://github.com/embrace-io/embrace-react-native-sdk/pull/51
  * main implementation is found [here](https://github.com/embrace-io/embrace-react-native-sdk/tree/8bf8402c4fcdd61dbd5ddcedc1f1530afba0b801/packages/react-native-tracer-provider/src)


packages/react-native-tracer-provider/native-src/android/app/dependencies.gradle
```diff
@@ -0,0 +1,5 @@
+dependencies {
+  // TODO keep this in sync with the version from the core/ package
+  api "io.embrace:embrace-android-sdk:6.9.1"
```
fnewberg, line 3: would it make sense to have some tooling around bumping the android SDK version to keep them all in sync?  if we don't have tooling i would do the following:

* add a comment like this one in `packages/core/android/gradle.properties`
* mention all places that need to be changed in `DEVELOPING.md`

jpmunz: I got blocked on what I was working on so tackled this instead, here's what I was going to go with: https://github.com/embrace-io/embrace-react-native-sdk/pull/55

packages/react-native-tracer-provider/native-src/android/app/src/main/java/io/embrace/reactnativetracerprovider/ReactNativeTracerProviderModule.kt
```diff
@@ -0,0 +1,337 @@
...
+            log.warning("could not retrieve span with id: $id")
+        }
+        return span
+    }
+
+    /**
+     * Methods to allow the JS side to conform to @opentelemetry-js/api
+     */
+
+    @ReactMethod fun getTracer(name: String, version: String, schemaUrl: String) {
```
priettt, line 178: Annotations are usually placed above the function, like this:

```
@ReactMethod
fun getTracer
```

priettt, line 178: maybe we should rename this to buildTracer or something similar? as we aren't really returning the tracer here

priettt (approved):
👍

packages/react-native-tracer-provider/native-src/android/app/src/main/java/io/embrace/reactnativetracerprovider/ReactNativeTracerProviderModule.kt
jpmunz: updated

jpmunz: yep good point, I had called this `getTracer` originally to match the interface name on the JS side but it is misleading, renamed to `setupTracer`

jpmunz:
@embrace-io/opensource-react-native  not fully reviewed yet but I'm going to merge this into the `feature-react-native-tracer-provider` feature branch as is to make development on the iOS side easier, the feature branch was on pause long enough that it should probably be fully re-reviewed regardless

PR #56, opened 2024-07-16
iOS 6.0 Migration

TheMutsi:
## Goal

Create Swift class to use the new methods from iOS 6.0

### DIFF
--- a/packages/core/ios/RNEmbrace/EmbraceManager.swift
+++ b/packages/core/ios/RNEmbrace/EmbraceManager.swift
@@ -21,7 +21,7 @@ class EmbraceManager: NSObject {
     @objc(setJavaScriptBundlePath:resolver:rejecter:)
     func setJavaScriptBundlePath(_ path: String, resolver resolve: @escaping RCTPromiseResolveBlock, rejecter reject: @escaping RCTPromiseRejectBlock) {
         do {
-            try Embrace.client?.metadata.addResource(key: EmbraceKeys.javaScriptBundleURL.rawValue, value: path)
+            try Embrace.client?.metadata.addResource(key: EmbraceKeys.javaScriptBundleURL.rawValue, value: path, .process)
             resolve(true)
         }  catch let error {
             reject("SET_JS_BUNDLE_PATH_ERROR", "Error setting JavaScript bundle path", error)
@@ -110,7 +110,7 @@ class EmbraceManager: NSObject {
     @objc(setReactNativeSDKVersion:resolver:rejecter:)
     func setReactNativeSDKVersion(_ version: String, resolver resolve: @escaping RCTPromiseResolveBlock, rejecter reject: @escaping RCTPromiseRejectBlock) {
         do {
-            try Embrace.client?.metadata.addResource(key: EmbraceKeys.embraceReactNativeSdkVersion.rawValue, value: version)
+            try Embrace.client?.metadata.addResource(key: EmbraceKeys.embraceReactNativeSdkVersion.rawValue, value: version, .process)
             resolve(true)
         } catch let error {
             reject("SET_RN_SDK_VERSION", "Error setting ReactNative SDK version", error)
@@ -133,7 +133,7 @@ class EmbraceManager: NSObject {
     @objc(setJavaScriptPatchNumber:resolver:rejecter:)
     func setJavaScriptPatchNumber(_ patch: String, resolver resolve: @escaping RCTPromiseResolveBlock, rejecter reject: @escaping RCTPromiseRejectBlock) {
         do {
-            try Embrace.client?.metadata.addResource(key: EmbraceKeys.javaScriptPatchNumber.rawValue, value: patch)
+            try Embrace.client?.metadata.addResource(key: EmbraceKeys.javaScriptPatchNumber.rawValue, value: patch, .process)
             resolve(true)
         } catch let error {
             reject("SET_JAVASCRIPT_PATCH_NUMBER", "Error setting JavasScript Patch Number", error)
@@ -171,7 +171,7 @@ class EmbraceManager: NSObject {
                 return
             }
             
-            try Embrace.client?.metadata.addResource(key: EmbraceKeys.javaScriptBundleURL.rawValue, value: url.path)
+            try Embrace.client?.metadata.addResource(key: EmbraceKeys.javaScriptBundleURL.rawValue, value: url.path, .process)
             resolve(true)
             
 #else
@@ -205,7 +205,7 @@ class EmbraceManager: NSObject {
     @objc(setReactNativeVersion:resolver:rejecter:)
     func setReactNativeVersion(_ version: String, resolver resolve: @escaping RCTPromiseResolveBlock, rejecter reject: @escaping RCTPromiseRejectBlock) {
         do {
-            try Embrace.client?.metadata.addResource(key: EmbraceKeys.reactNativeVersion.rawValue, value: version)
+            try Embrace.client?.metadata.addResource(key: EmbraceKeys.reactNativeVersion.rawValue, value: version, .process)
             resolve(true)
         } catch let error {
             reject("SET_RECT_NATIVE_VERSION", "Error setting React Native Number", error)
--- a/packages/core/ios/RNEmbrace/EmbraceManager.swift
+++ b/packages/core/ios/RNEmbrace/EmbraceManager.swift
@@ -1,16 +1,17 @@
 import Foundation
 import EmbraceIO
-import React 
+import React
+import EmbraceCommon
 
 #if canImport(CodePush)
 import CodePush
 #endif
 
 enum EmbraceKeys: String {
-    case reactNativeVersion = "io.embrace.reactnative.version"
-    case embraceReactNativeSdkVersion = "io.embrace.reactnative.sdk.version"
-    case javaScriptPatchNumber = "io.embrace.javascript.patch"
-    case javaScriptBundleURL = "io.embrace.jsbundle.url"
+  case reactNativeVersion = "io.embrace.reactnative.version"
+  case embraceReactNativeSdkVersion = "io.embrace.reactnative.sdk.version"
+  case javaScriptPatchNumber = "io.embrace.javascript.patch"
+  case javaScriptBundleURL = "io.embrace.jsbundle.url"
 }
 
 @objc(EmbraceManager)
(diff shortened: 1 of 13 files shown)


packages/core/ios/RNEmbrace/EmbraceManager.swift
```diff
@@ -0,0 +1,301 @@
...
+    do {
+      try Embrace.client?.metadata.add(persona: PersonaTag(persona), lifespan: .session)
+      resolve(true)
+    } catch let error {
+      reject("ADD_USER_PERSOMAS", "Error adding an User Personas", error)
+    }
+  }
+  
+  @objc(clearUserPersonas:resolver:rejecter:)
+  func clearUserPersonas(_ persona: String, resolve: @escaping RCTPromiseResolveBlock, rejecter reject: @escaping RCTPromiseRejectBlock) {
```
jpmunz, line 195: looks like a duplicate of `clearUserPersona`, can it be removed?

packages/core/ios/RNEmbrace/EmbraceManager.swift
```diff
@@ -0,0 +1,301 @@
...
+  case reactNativeVersion = "io.embrace.reactnative.version"
+  case embraceReactNativeSdkVersion = "io.embrace.reactnative.sdk.version"
+  case javaScriptPatchNumber = "io.embrace.javascript.patch"
+  case javaScriptBundleURL = "io.embrace.jsbundle.url"
+}
+
+@objc(EmbraceManager)
+class EmbraceManager: NSObject {
+  @objc
+  func setJavaScriptBundlePath(_ path: String, resolver resolve: @escaping RCTPromiseResolveBlock, rejecter reject: @escaping RCTPromiseRejectBlock) {
```
jpmunz, line 21: for this and the other methods I think `@escaping` is only required when using the variables in a closer, are they needed here because of the `do...catch`? Are they needed in other methods that don't include that?

packages/core/ios/RNEmbrace/EmbraceManager.swift
```diff
@@ -0,0 +1,301 @@
...
+  @objc
+  func isStarted(_ resolve: @escaping RCTPromiseResolveBlock, rejecter reject: @escaping RCTPromiseRejectBlock) {
+    if let embraceStarted = Embrace.client?.started {
+      resolve(embraceStarted)
+    } else {
+      resolve(false)
+    }
+  }
+  
+  @objc(startNativeEmbraceSDK:resolver:rejecter:)
```
jpmunz, line 39: some of the `@objc` decorators have the method name + args but others are just the decorator, in what cases are they are required? Should we just always have them?

packages/core/ios/RNEmbrace/EmbraceManager.swift
```diff
@@ -0,0 +1,301 @@
...
+      try Embrace.client?.metadata.addResource(key: EmbraceKeys.embraceReactNativeSdkVersion.rawValue, value: version)
+      resolve(true)
+    } catch let error {
+      reject("SET_RN_SDK_VERSION", "Error setting ReactNative SDK version", error)
+    }
+  }
+  
+  
+  @objc
+  func setUserName(_ userName: String, resolver resolve: @escaping RCTPromiseResolveBlock, rejecter reject: @escaping RCTPromiseRejectBlock) {
```
jpmunz, line 121:
```suggestion
  func setUsername(_ userName: String, resolver resolve: @escaping RCTPromiseResolveBlock, rejecter reject: @escaping RCTPromiseRejectBlock) {
```

packages/core/ios/RNEmbrace/EmbraceManager.swift
```diff
@@ -0,0 +1,301 @@
...
+    rejecter reject: @escaping RCTPromiseRejectBlock
+  ) {
+    do {
+#if canImport(CodePush)
+      guard let url = try CodePush.bundleURL() else {
+        reject("GET_CODEPUSH_BUNDLE_URL", "Error getting Codepush Bundle URL", nil)
+        return
+      }
+      
+      try Embrace.client?.metadata.addResource(key: EmbraceKeys.javaScriptBundleURL.rawValue, value: url.path)
```
jpmunz, line 173: just confirming this is meant to use the same key as `setJavaScriptBundlePath`

packages/core/ios/RNEmbrace/EmbraceManager.swift
```diff
@@ -0,0 +1,301 @@
...
+    }
+  }
+  
+  @objc(addUserPersona:resolver:rejecter:)
+  func addUserPersona(_ persona: String, resolve: @escaping RCTPromiseResolveBlock, rejecter reject: @escaping RCTPromiseRejectBlock) {
+    do {
+      try Embrace.client?.metadata.add(persona: PersonaTag(persona), lifespan: .session)
+      resolve(true)
+    } catch let error {
+      reject("ADD_USER_PERSOMAS", "Error adding an User Personas", error)
```
jpmunz, line 190:
```suggestion
      reject("ADD_USER_PERSONA", "Error adding User Persona", error)
```

packages/core/ios/RNEmbrace/EmbraceManager.swift
```diff
@@ -0,0 +1,301 @@
...
+#endif
+    } catch let error {
+      reject("CHECK_AND_SET_CODEPUSH_BUNDLE_URL", "Error setting Codepush Bundle URL", error)
+    }
+  }
+  
+  @objc(addUserPersona:resolver:rejecter:)
+  func addUserPersona(_ persona: String, resolve: @escaping RCTPromiseResolveBlock, rejecter reject: @escaping RCTPromiseRejectBlock) {
+    do {
+      try Embrace.client?.metadata.add(persona: PersonaTag(persona), lifespan: .session)
```
jpmunz, line 187: is the casting of `PersonaTag` needed here? From the iOS docs seems like you could just pass a string: https://embrace.io/docs/ios/open-source/metadata-personas/#adding-user-personas-in-the-embrace-sdk

packages/core/ios/RNEmbrace/EmbraceManager.swift
```diff
@@ -0,0 +1,301 @@
...
+    }
+  }
+  
+  @objc
+  func clearAllUserPersonas(_ resolve: @escaping RCTPromiseResolveBlock, rejecter reject: @escaping RCTPromiseRejectBlock) {
+    do {
+      try Embrace.client?.metadata.removeAllPersonas()
+      resolve(true)
+    } catch let error {
+      reject("CLEAR_ALL_USER_PERSOMAS", "Error clearing all User Personas", error)
```
jpmunz, line 210:
```suggestion
      reject("CLEAR_ALL_USER_PERSONAS", "Error clearing all User Personas", error)
```

packages/core/ios/RNEmbrace/EmbraceManager.swift
```diff
@@ -0,0 +1,301 @@
...
+    }
+  }
+  
+  @objc
+  func setReactNativeVersion(_ version: String, resolver resolve: @escaping RCTPromiseResolveBlock, rejecter reject: @escaping RCTPromiseRejectBlock) {
+    do {
+      try Embrace.client?.metadata.addResource(key: EmbraceKeys.reactNativeVersion.rawValue, value: version)
+      resolve(true)
+    } catch let error {
+      reject("SET_RECT_NATIVE_VERSION", "Error setting React Native Number", error)
```
jpmunz, line 220:
```suggestion
      reject("SET_REACT_NATIVE_VERSION", "Error setting React Native Version", error)
```

packages/core/ios/RNEmbrace/EmbraceManager.swift
```diff
@@ -0,0 +1,301 @@
...
+    }
+  }
+  
+  @objc
+  func removeSessionProperty(_ key: String, resolver resolve: @escaping RCTPromiseResolveBlock, rejecter reject: @escaping RCTPromiseRejectBlock) {
+    do {
+      // TODO REfactor to include lifespan
```
jpmunz, line 227: this comment still relevant?

packages/core/ios/RNEmbrace/EmbraceManager.swift
```diff
@@ -0,0 +1,301 @@
...
+  @objc
+  func clearUserEmail(_ resolve: @escaping RCTPromiseResolveBlock, rejecter reject: @escaping RCTPromiseRejectBlock) {
+    Embrace.client?.metadata.userEmail = nil
+    resolve(true)
+  }
+  
+  @objc
+  func setJavaScriptPatchNumber(_ patch: String, resolver resolve: @escaping RCTPromiseResolveBlock, rejecter reject: @escaping RCTPromiseRejectBlock) {
+    do {
+      try Embrace.client?.metadata.addResource(key: EmbraceKeys.javaScriptPatchNumber.rawValue, value: patch)
```
jpmunz, line 135: for most of these utility ones we should be setting lifespan as `.process`, see https://embrace-io.slack.com/archives/C07CSGLSARJ/p1721928160750769

packages/core/ios/RNEmbrace/EmbraceManager.swift
```diff
@@ -0,0 +1,301 @@
...
+      resolve(true)
+    } catch let error {
+      reject("ADD_SESSION_PROPERTY", "Error adding Session Property", error)
+    }
+  }
+  
+  @objc(clearUserPersona:resolver:rejecter:)
+  func clearUserPersona(_ persona:String, resolve: @escaping RCTPromiseResolveBlock, rejecter reject: @escaping RCTPromiseRejectBlock) {
+    do {
+      try Embrace.client?.metadata.remove(persona: PersonaTag(persona), lifespan: .session)
```
jpmunz, line 261: same here for removing the `PersonaTag` casting

packages/core/ios/RNEmbrace/EmbraceManager.swift
```diff
@@ -0,0 +1,301 @@
...
+    switch inputString {
+    case "info":
+      return .info
+    case "warning":
+      return .warn
+    default:
+      return .error
+    }
+  }
+}
```
jpmunz, line 301: for this file I'd suggest reorganizing the methods so that related ones are together, I'd say we can just follow how the table is laid out in: https://www.notion.so/embraceio/Deprecation-of-Embrace-specific-methods-in-the-RN-API-in-favour-of-OTEL-equivalents-62a6fa6d3e344490809b156fad2938e7?pvs=4 with comment blocks separating the categories like "Utility" and "Session"

packages/core/src/index.ts
```diff
@@ -95,18 +99,19 @@ export const initialize = async ({
...
-        message = unhandledPromiseRejectionPrefix + error.message;
-        st = error.stack || "";
-      }
-      return NativeModules.EmbraceManager.logMessageWithSeverityAndProperties(
-        message,
-        ERROR,
-        {},
-        st,
-      );
+      // TODO REFACTOR WHEN iOS IMPLEMENT THE METHOD
```
jpmunz, line 102: can this be uncommented now that we have `logMessageWithSeverityAndProperties` implemented on the Swift side?

packages/core/src/index.ts
```diff
@@ -130,10 +135,13 @@ const buildVersionStr = ({
 };
 
 export const endAppStartup = (properties?: Properties): Promise<boolean> => {
```
jpmunz, line 137: `endAppStartup` can be removed, see https://embrace-io.slack.com/archives/C07CSGLSARJ/p1721665753052099

packages/core/src/index.ts
```diff
@@ -170,55 +178,12 @@ export const logScreen = (screenName: string): Promise<boolean> => {
...
-      name,
-      null,
-      properties,
-    );
-  } else {
-    return NativeModules.EmbraceManager.startMomentWithName(name);
-  }
-};
+export const addUserPersona = (persona: string): Promise<boolean> => {
+  // TODO REFACTOR WHEN iOS IMPLEMENT THE METHOD
```
jpmunz, line 182: can uncomment I believe

packages/core/src/index.ts
```diff
@@ -348,17 +336,22 @@ export const logNetworkClientError = (
...
+  //   startInMillis,
+  //   endInMillis,
+  //   errorType,
+  //   errorMessage,
+  // );
+  return createFalsePromise();
 };
 export const getLastRunEndState = (): Promise<SessionStatus> =>
-  NativeModules.EmbraceManager.getLastRunEndState();
+  // TODO REFACTOR WHEN iOS IMPLEMENT THE METHOD
```
jpmunz, line 352: can be uncommented

packages/core/ios/RNEmbrace/EmbraceManager.swift
```diff
@@ -0,0 +1,301 @@
+import Foundation
+import EmbraceIO
+import EmbraceCore
+import EmbraceCrash
+import EmbraceCommonInternal
```
ArielDemarco, line 5: Is necessary to import `EmbraceCommonInternal`? Which is the object/class/method needing this "dependency" to be imported?

packages/core/ios/RNEmbrace/EmbraceManager.swift
```diff
@@ -0,0 +1,301 @@
...
+    if let sessionId = Embrace.client?.currentSessionId() {
+      resolve(sessionId)
+    } else {
+      reject("GET_SESSION_ID", "Error getting sessionId", nil)
+    }
+  }
+  
+  @objc
+  func addBreadcrumb(_ event: String, resolver resolve: @escaping RCTPromiseResolveBlock, rejecter reject: @escaping RCTPromiseRejectBlock) {
+    //TODO Refactor SPAN
```
ArielDemarco, line 91: This should be something like
```swift
Embrace.client?.add(.breadcrumb(event)
resolve(true)
```

packages/core/ios/RNEmbrace/EmbraceManager.swift
jpmunz: this might be fixed but it was needed for the `LogSeverity` enum

packages/core/ios/RNEmbrace/EmbraceManager.swift
jpmunz: noted in https://www.notion.so/embraceio/Implement-addBreadcrumb-in-iOS-native-module-8ec0d6f2e04d4b49a7b8fa7d669cf59b?pvs=4, can do it here if it's quick or punt to that ticket in the interest of wrapping up this PR

packages/core/src/index.ts
```diff
@@ -304,19 +277,31 @@ export const removeSessionProperty = (key: string) => {
 };
 
 export const getSessionProperties = () => {
```
jpmunz, line 279: `getSessionProperties` can be removed from here and the Android native module

packages/core/ios/RNEmbrace/EmbraceManager.swift
ArielDemarco: Just providing context; if you guys prefer to do it in another PR, go ahead!

packages/core/src/index.ts
```diff
@@ -304,19 +277,31 @@ export const removeSessionProperty = (key: string) => {
...
+  // TODO REFACTOR WHEN iOS IMPLEMENT THE METHOD
+
+  // return NativeModules.EmbraceManager.getSessionProperties();
+
+  return createFalsePromise();
 };
 
 export const endSession = (clearUserInfo: boolean = false) => {
-  return NativeModules.EmbraceManager.endSession(clearUserInfo);
+  return NativeModules.EmbraceManager.endSession();
```
jpmunz, line 288: we need to continue to pass `clearUserInfo` through here since the Android native module expects it, we'll just have to ignore on the iOS side

packages/core/ios/RNEmbrace/EmbraceManager.swift
TheMutsi: When you have a parameter else than the promises, you need to specify it

packages/core/ios/RNEmbrace/EmbraceManager.swift
TheMutsi: To change this I need to change the android one. Do you want me to change both ?

packages/core/ios/RNEmbrace/EmbraceManager.swift
TheMutsi: Yes

packages/core/ios/RNEmbrace/EmbraceManager.swift
TheMutsi: Yes, since we need to refactor to include the other lifespan. Today we only support 1

packages/core/ios/RNEmbrace/EmbraceManager.swift
jpmunz: that seems inconsistent with the current methods though, e.g. `setUserIdentifier` below has the `userIdentifier` parameter but it only has the `@objc` decorator by itself?

packages/core/src/index.ts
TheMutsi: mmm in my local I dont have that parameter, maybe this PR is not up to date

Anyway I will push right now the test fix

I removed the parameter `clearUserInfo: boolean = false`. since android expose 2 methods, and I will use the first method, without parameters to have the same method signature and functionality than iOS

Android SDK
```
    @Override
    public synchronized void endSession() {
        endSession(false);
    }
```

packages/core/src/index.ts
TheMutsi: I slacked you, because I think we need to recreate it for ios

packages/core/src/index.ts
TheMutsi: I am working in other branch, let it commented by now

packages/core/ios/RNEmbrace/EmbraceManager.swift
TheMutsi: Yes

packages/core/ios/RNEmbrace/EmbraceManager.swift
TheMutsi: Since we are using promise in RN, it's good practice to use @escaping when you work with async functions. Even though the content is not directly async right now. 
If the content is not async, you could remove it. Let me know if you want me to remove it ( I would keep it)

jpmunz: ok let's keep it

jpmunz (approved):
looks good, a few remaining issues:
* endSession needs to no longer take `clearUserInfo` on the Android side
* typo from: https://github.com/embrace-io/embrace-react-native-sdk/pull/56#discussion_r1691877458
* failing check

PR #59, opened 2024-07-22
EMBR-4422 native module interface for spans

jpmunz:
Adding in the iOS native module methods that we previously had for the spans API, these depend on a spans repository for keeping track of spans by ID following a similar implementation as [Android's](https://github.com/embrace-io/embrace-android-sdk/blob/master/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/SpanRepository.kt).

Need to integration test this using the example app but the code should be good to review as is

Future:
* Bridge methods are similar to the [trace provider bridge](https://github.com/embrace-io/embrace-react-native-sdk/compare/feature-react-native-tracer-provider...jpmunz/EMBR-3908-ios-tracer-provider?expand=1), should see if code can be re-used, particularly for the span repository


packages/core/RNEmbrace.podspec
```diff
@@ -9,8 +9,9 @@ Pod::Spec.new do |s|
   s.license      = package["license"]
   s.authors      = package["author"]
 
-  s.ios.deployment_target = '9.0'
-  s.tvos.deployment_target = '10.0'
+  s.ios.deployment_target = '13.0'
```
jpmunz, line 12: are these correct?

packages/core/test-project/ios/RNEmbraceTests/RNEmbraceTests.swift
jpmunz, line 1: Note: adding these tests were helpful for me during development but since we haven't previously had iOS tests these are not yet hooked up the CI, I punted that to a separate ticket which I'll try and get to but shouldn't block release of iOS 6 migration we'll just have to keep running manuallyL https://www.notion.so/embraceio/Hook-Swift-tests-up-to-CI-4f67687617de471f8c71b7d9de4b941a?pvs=4

packages/core/test-project/ios/RNEmbraceTests/RNEmbraceTests.swift
```diff
@@ -1,38 +1,515 @@
...
+        XCTAssertTrue(exportedSpans[1].hasEnded)
+        XCTAssertEqual(exportedSpans[2].name, "my-span-invalid-parent")
+        XCTAssertNil(exportedSpans[2].parentSpanId)
+        XCTAssertEqual(exportedSpans[2].attributes.count, 1)
+        XCTAssertEqual(exportedSpans[2].attributes["emb.type"]!.description, "perf")
+        XCTAssertEqual(exportedSpans[2].events.count, 0)
+        XCTAssertTrue(exportedSpans[2].hasEnded)
+    }
+
+    // TODO fails on 6.2 currently
```
jpmunz, line 424: waiting on fix

packages/core/RNEmbrace.podspec
fractalwrench: Yes - confirmed in internal slack

packages/core/ios/RNEmbrace/EmbraceManager.m
```diff
@@ -84,6 +84,41 @@ @interface RCT_EXTERN_MODULE(EmbraceManager, NSObject)
                   resolver:(RCTPromiseResolveBlock)resolve
                   rejecter:(RCTPromiseRejectBlock)reject)
 
+RCT_EXTERN_METHOD(startSpan:(NSString *)name
+                  parentSpanId:(NSString *)parentSpanId
+                  startTimeMS:(nonnull NSNumber *)startTimeMS
```
fractalwrench, line 89: minor nit: could this be `startTimeMs`? If that's inconsistent with how naming has been done then it's not that important

fractalwrench (approved):
LGTM

packages/core/ios/RNEmbrace/EmbraceManager.m
jpmunz: yep let me update

packages/core/ios/RNEmbrace/SpanRepository.swift
```diff
@@ -0,0 +1,90 @@
...
+    private var activeSpans = [String: Span]()
+    private var completedSpans = [String: Span]()
+    private var log = OSLog(subsystem: "Embrace", category: "SpanRepository")
+
+    init() {
+        NotificationCenter.default.addObserver(
+            self,
+            selector: #selector(onSessionEnded),
+            // TODO possible to reference this constant? https://github.com/embrace-io/embrace-apple-sdk/blob/main/Sources/EmbraceCore/Session/SessionController.swift#L13
+            name: Notification.Name("embrace.session.will_end"),
```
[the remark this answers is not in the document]
ArielDemarco: We're in the process of exposing a new module with our constants so this becomes easier. IDK in particular if this one, but we can discuss it.
That being said, this shouldn't be a blocker to merge

packages/core/ios/RNEmbrace/SpanRepository.swift
```diff
@@ -0,0 +1,90 @@
...
+            }
+        }
+
+        return span
+    }
+
+    func spanStarted(span: Span) -> String {
+        let key = getKey(span)
+
+        // TODO does concurrent access matter for grabbing count?
```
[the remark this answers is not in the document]
ArielDemarco: No it doesn't at all. Even though having the actual exact number is ok, we apply limits to OTel stuff mostly to prevent any problems on the SDKs (like growing memory too much, or using too much disk space). Actual limits (if we'd charge by X amount of spans/events/etc per session) should be also covered in the backend.
FWIW, on the SDK side, we have an in-memory counter so we don't go constantly to the database for this validation. And even though you can do this in a "thread-safe" manner (having a sync queue / lock accessing and setting the actual number of `activeSpans.count` to have the exact number) I believe this is not necessary.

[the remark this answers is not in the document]
jpmunz: sounds good, I'll leave as is then

PR #61, opened 2024-07-30
iOS 6 - addBreadcrumb - logRNAction

TheMutsi:
## Goal

Refactor RN Action to use Spans

Add Breadcrumb in ios


packages/action-tracker/src/index.ts
```diff
@@ -15,14 +15,29 @@ import {
...
       if (!next || typeof next !== "function") {
         console.warn(
           "[Embrace] The state managment middleware was not applied succesfully",
         );
         return 0;
       }
       const startTime = new Date().getTime();
+      const attributes = {
+        name: action.type.toString().toUpperCase(),
+        "emb.send_immediately": true,
```
jpmunz, line 28: discussed offline (https://embrace-io.slack.com/archives/C05NTFHCFBJ/p1721164675230939), we can remove the  "emb.send_immediately" attribute

packages/action-tracker/src/index.ts
```diff
@@ -32,13 +47,23 @@ export const buildEmbraceMiddleware = () => {
...
+          attributePromises.push(
+            NativeModules.EmbraceManager.addSpanAttributeToSpan(
+              spanId,
+              "outcome",
+              "SUCCESS",
+            ),
+          );
+
+          await Promise.all(attributePromises);
+          NativeModules.EmbraceManager.recordCompletedSpan(
```
jpmunz, line 59: should be `stopSpan` here, `recordCompletedSpan` wouldn't close the one that was started above

packages/action-tracker/src/index.ts
```diff
@@ -32,13 +47,23 @@ export const buildEmbraceMiddleware = () => {
...
-            action.type.toString().toUpperCase(),
+          attributePromises.push(
+            NativeModules.EmbraceManager.addSpanAttributeToSpan(
+              spanId,
+              "outcome",
+              "SUCCESS",
+            ),
+          );
+
+          await Promise.all(attributePromises);
```
jpmunz, line 58: I think we're better off just firing these off and not waiting on them and if something does go wrong we wouldn't want to block the app because of this instrumentation

packages/action-tracker/src/index.ts
TheMutsi: Yes this was the example that I gave you through slack, I reverted it to let stopSpan

packages/action-tracker/src/index.ts
TheMutsi: but, If we do that, we cannot guarantee that the stop span didnt finish first than the other promises. Can a stopped span received new attributes ?

packages/action-tracker/src/index.ts
TheMutsi: I removed that attribute. What do you want to discuss offline ?  ( I cannot enter to that link)

jpmunz: just that `emb.send_immediately` was an internal attribute set by the Android SDK and not something we need to worry about when using the spans API here

packages/action-tracker/src/index.ts
jpmunz: nope a stopped span becomes readonly so new attributes would be ignored

I think it's an unlikely enough case that I'd be fine with the simpler fire and forget approach here, but if we are worried about it I should confirm my assumption that this would actually have an impact on the user's code. Do you know how Redux's middleware management would handle this having an async action? Would it wait on it or would it just continue on to the next middleware?

packages/core/ios/RNEmbrace/EmbraceManager.swift
```diff
@@ -98,13 +98,12 @@ class EmbraceManager: NSObject {
...
     @objc(addBreadcrumb:resolver:rejecter:)
     func addBreadcrumb(_ event: String, resolver resolve: @escaping RCTPromiseResolveBlock, rejecter reject: @escaping RCTPromiseRejectBlock) {
-        //TODO Refactor SPAN
-        //      do {
-        //        try Embrace.client?.add(event: event)
-        //          resolve(true)
-        //      } catch {
-        //          reject(false)
-        //      }
+        // Add function returns empty if it sucess, so I specify true as return
```
jpmunz, line 101:
```suggestion
        // Add function returns empty if it succeeds, so I specify true as return
```

packages/action-tracker/src/index.ts
jpmunz: rest of the changes lgtm after testing and resolving this ^ question

packages/action-tracker/src/index.ts
```diff
@@ -15,14 +15,29 @@ import {
...
+      const attributes = {
+        name: action.type.toString().toUpperCase(),
+        payload_size: zip(action.payload || 0).length,
+        "emb.type": "sys.rn_action",
+        outcome: "INCOMPLETE",
+      };
+      const spanId = await NativeModules.EmbraceManager.startSpan(
+        "emb-rn-action",
+        startTime,
+      );
```
jpmunz, line 35: requires a second parameter for parentSpanId

```suggestion
      const spanId = await NativeModules.EmbraceManager.startSpan(
        "emb-rn-action",
        undefined,
        startTime,
      );
```

packages/action-tracker/src/index.ts
```diff
@@ -32,14 +47,17 @@ export const buildEmbraceMiddleware = () => {
...
+            NativeModules.EmbraceManager.addSpanAttributeToSpan(
+              spanId,
+              "outcome",
+              "SUCCESS",
+            ),
           );
+
+          await Promise.all(attributePromises);
+
+          NativeModules.EmbraceManager.stopSpan(spanId, undefined, endTime);
```
jpmunz, line 60:
```suggestion
          NativeModules.EmbraceManager.stopSpan(spanId, "None", endTime);
```

packages/action-tracker/src/index.ts
```diff
@@ -15,14 +15,29 @@ import {
...
       if (!next || typeof next !== "function") {
         console.warn(
           "[Embrace] The state managment middleware was not applied succesfully",
         );
         return 0;
       }
       const startTime = new Date().getTime();
+      const attributes = {
+        name: action.type.toString().toUpperCase(),
+        payload_size: zip(action.payload || 0).length,
```
jpmunz, line 28: attributes must be strings, I was seeing this show up in the dashboard correctly after casting:

```suggestion
        payload_size: (zip(action.payload || 0).length).toString(),
```

packages/action-tracker/src/index.ts
```diff
@@ -15,14 +15,30 @@ import {
...
+        payload_size: zip(action.payload || 0).length.toString(),
+        "emb.type": "sys.rn_action",
+        outcome: "INCOMPLETE",
+      };
+      const spanId = await NativeModules.EmbraceManager.startSpan(
+        "emb-rn-action",
+        undefined,
+        startTime,
+      );
+      const attributePromises: Promise<boolean>[] = Object.entries(
```
jpmunz, line 37: seems `attributePromises` is not needed anymore? Can the calls to `addSpanAttributeToSpan` just be made without adding the result to the list?

packages/core/src/__tests__/embrace.test.ts
```diff
@@ -332,14 +338,21 @@ describe("Session Properties Tests", () => {
 
 describe("Payers Test", () => {
   test("setUserAsPayer", async () => {
-    await setUserAsPayer();
-    // TODO uncomment the expect once the method is implemented
-    // expect(mockSetUserAsPayer).toHaveBeenCalled();
+    const promiseToResolve = setUserAsPayer();
+    jest.runAllTimers();
```
jpmunz, line 342: `jest.runAllTimers();` calls shouldn't be needed here

packages/core/src/__tests__/embrace.test.ts
```diff
@@ -332,14 +338,21 @@ describe("Session Properties Tests", () => {
...
+    const result = await promiseToResolve;
+
+    expect(mockSetUserAsPayer).toHaveBeenCalled();
+    expect(result).toBe(false);
```
jpmunz, line 346: there's not much value in asserting this since it's just the value returned from the mock in this test file

jpmunz (approved):

PR #62, opened 2024-07-30
Update iOS initialization to configure from code instead of plist

jpmunz:
Documentation changes in progress separately

Updating the JS side to accept a configuration object that for now is just used for iOS setup and passing that through to the swift code. Only a subset of configuration is exposed to get feature parity with what we allowed to be configured in 5.x


packages/core/ios/RNEmbrace/EmbraceManager.swift
```diff
@@ -22,6 +23,34 @@ private let EVENT_NAME_KEY = "name"
...
+                } else if key == "disableCrashReporter", let value = value as? Bool {
+                    disableCrashReporter = value
+                } else if key == "disableAutomaticViewCapture", let value = value as? Bool {
+                    disableAutomaticViewCapture = value
+                } else if key == "endpointBaseUrl", let value = value as? String {
+                    endpointBaseUrl = value
+                }
+            }
+        }
+    }
```
TheMutsi, line 50: I dont see any advantage to use for if we are going to check with if each variable inside. Maybe just decode the json with JSONDecoder (but it will fail if there is a typo error, you could handle the error)

or just something like this

```self.appId = from["appId"] as? String ?? ""
  self.appGroupId = from["appGroupId"] as? String
  self.disableCrashReporter = from["disableCrashReporter"] as? Bool ?? false
  self.disableAutomaticViewCapture = from["disableAutomaticViewCapture"] as? Bool ?? false
  self.endpointBaseUrl = from["endpointBaseUrl"] as? String```

File: packages/core/src/interfaces/Config.ts [hunk 2]
```diff
@@ -0,0 +1,12 @@
+/**
+ * Defines the SDK configuration that can be passed into `initialize`
+ */
+export interface SDKConfig {
+  ios?: {
+    appId: string;
+    appGroupId?: string;
+    disableCrashReporter?: boolean;
```
TheMutsi, line 8: mmm shouldn't be there other attribute to set other crash reporter than us? like crashlitics or something like that, or that is managed in other place

TheMutsi (approved):
lgtm

packages/core/src/interfaces/Config.ts
jpmunz: this was to keep feature parity for what we currently talk about in the docs: https://embrace.io/docs/react-native/integration/add-embrace-sdk/

so allowing the user to just disable our crash reporter. Automatically configuring crashlytics in place of ours would be more complicated and something we decided not to provide for now, a customer would have to do this themselves in objective-c / swift

PR #63, opened 2024-08-01
EMBR-4337 update the install/uninstall scripts for iOS 6

jpmunz:
Few different changes here:
- don't add the plist on install
- add a common EmbraceInitializer.swift file, 2 cases for importing it, objective-c app delegate vs. swift app delegate
	- if it's objective-c need to add a bridging header file if there isn't one already
- keep the plist and old import removals in the uninstall script in case it gets run on an old version

Manual steps are documented in: https://github.com/embrace-io/embrace-docs/pull/596


packages/core/scripts/postlink/index.ts
jpmunz, line 1: @TheMutsi  along with  setup/run.ts, setup/setup.ts, setup/runUpdate.ts I also had to delete this file since it was referencing setup/setup.ts, it didn't seem to be used but let me know if was needed somewhere

packages/core/scripts/setup/ios.ts
```diff
@@ -135,74 +140,57 @@ export const addUploadBuildPhase = {
...
-    <string>${iosAppIDValue}</string>
-    <key>CRASH_REPORT_ENABLED</key>
-    <true/>
-  </dict>
-</plist>`;
+const getEmbraceInitializerContents = (appId: string) => {
+  return `import Foundation
+import EmbraceIO
+
+@objcMembers class EmbraceInitializer: NSObject {
```
jpmunz, line 177: @ArielDemarco this block shows the contents of the `EmbraceInitializer.swift` file that the install script will automatically add to the app's ios project

facostaembrace (approved):

PR #67, opened 2024-08-07
EMBR-4627 fix flaky tests

jpmunz:
Seeing a test failure on GHA that I wasn't seeing locally, I couldn't figure out the exact root cause but I thought it might be related to some of tests mocking the `NativeModules` inline, I think jest hoists those calls to the top of the file so there might be some weirdness when they're run in parallel

I'll see if this passes, moved all mocking to happen once at the top of each test file

### DIFF
--- a/packages/core/src/__tests__/embrace.test.ts
+++ b/packages/core/src/__tests__/embrace.test.ts
@@ -137,10 +137,9 @@ jest.mock("react-native", () => ({
 
 const mockSt = "this is a fake stack trace";
 const mockGenerateStackTrace = jest.fn();
-mockGenerateStackTrace.mockReturnValue(mockSt);
 jest.mock("../utils/ErrorUtil", () => ({
   ...jest.requireActual("../utils/ErrorUtil"),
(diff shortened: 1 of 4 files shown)


packages/core/src/__tests__/embrace.test.ts
```diff
@@ -1,204 +1,228 @@
...
-    const {clearUserIdentifier} = require("../index");
     await clearUserIdentifier();
-    expect(mock).toHaveBeenCalled();
+    expect(mockClearUserIdentifier).toHaveBeenCalled();
   });
 });
 
 describe("User Data Tests", () => {
+  const testUserId = "Lucia";
+  const testEmail = "lucia@nimble.la";
```
facostaembrace, line 164: I know you are not supposed to refactor, but maybe it worth to change this to something like test@test.com?

facostaembrace (approved):

packages/core/src/__tests__/embrace.test.ts
jpmunz: yep makes sense, updated

PR #69, opened 2024-08-09
Native start end view to track JS Screens

TheMutsi:
## Goal

Use spans for iOS and logRNAction for Android (This will be changed in the future) to track JS screens in react-native-navigation and react-navigation


packages/core/src/__tests__/embrace.test.ts
```diff
@@ -303,18 +303,53 @@ describe("Personas Tests", () => {
...
 
 describe("Custom Views Tests", () => {
-  const testView = "View";
-
   test("startView", async () => {
-    await startView(testView);
-    // TODO uncomment the expect once the method is implemented
-    // expect(mockStartView).toHaveBeenCalledWith(testView);
+    const mock = jest.fn();
+    jest.mock(
```
jpmunz, line 308: should avoid mocking these inline, there's a single NativeModules mock at the top that can be used, see https://github.com/embrace-io/embrace-react-native-sdk/pull/67

packages/core/ios/RNEmbrace/EmbraceManager.swift
```diff
@@ -329,6 +333,49 @@ class EmbraceManager: NSObject {
         
     }
     
+    @objc(startView:resolver:rejecter:)
+    func startView(_ viewName: String, resolver resolve: @escaping RCTPromiseResolveBlock, rejecter reject: @escaping RCTPromiseRejectBlock) {
+        let startTimeDate = Date()
```
jpmunz, line 338: I think this would be cleaner to call the `Embrace.client` methods for creating a span directly rather than the `startSpan` bridge methods which do some extra work, something like:
```
    let span = Embrace.client?.buildSpan(name: name).startSpan()
    // reject if span is nil

    span?.setAttribute(key: key, value: value)     // set ux.view, etc.
    Embrace.client?.flush(span!)

    let id = spanRepository.spanStarted(span: span!)
    // reject if id.isEmpty
    resolve(id)
```

packages/core/ios/RNEmbrace/EmbraceManager.swift
```diff
@@ -329,6 +333,49 @@ class EmbraceManager: NSObject {
...
+        }, rejecter: reject)
+        
+    }
+    
+    @objc(endView:resolver:rejecter:)
+    func endView(_ spanId: String, resolver resolve: @escaping RCTPromiseResolveBlock, rejecter reject: @escaping RCTPromiseRejectBlock) {
+        let endTimeDate = Date()
+        let endTimeMS = NSNumber(value: endTimeDate.timeIntervalSince1970 * 1000)
+        
+        stopSpan(spanId, errorCodeString: "", endTimeMs: endTimeMS, resolver: resolve, rejecter: reject)
```
jpmunz, line 376: this one is more direct but I think you could copy the relevant lines from `stopSpan` since you don't care about error code, end time

packages/core/ios/RNEmbrace/EmbraceManager.swift
jpmunz: actually the builder supports setting attributes so could save the flush call with something like:
```
let span = Embrace.client?.buildSpan(name: name)
                 .setAttribute(key:value:)
                 .setAttribute(key:value:)
                 .startSpan()
```

packages/core/ios/RNEmbrace/EmbraceManager.swift
TheMutsi: The same with start, if I do that I will rewrite the whole function

packages/core/src/__tests__/embrace.test.ts
```diff
@@ -303,18 +304,19 @@ describe("Personas Tests", () => {
...
 describe("Custom Views Tests", () => {
-  const testView = "View";
-
   test("startView", async () => {
-    await startView(testView);
-    // TODO uncomment the expect once the method is implemented
-    // expect(mockStartView).toHaveBeenCalledWith(testView);
+    const promiseToResolve = startView(testView);
+
+    jest.runAllTimers();
```
jpmunz, line 310: `jest.runAllTimers` calls in these shouldn't be needed

packages/core/ios/RNEmbrace/EmbraceManager.swift
```diff
@@ -20,565 +20,599 @@ private let EVENT_TIMESTAMP_KEY = "timeStampMs"
...
+  func clearUserPersona(_ persona:String, resolve: @escaping RCTPromiseResolveBlock, rejecter reject: @escaping RCTPromiseRejectBlock) {
+    do {
+      try Embrace.client?.metadata.remove(persona: PersonaTag(persona), lifespan: .session)
+      resolve(true)
+    }catch let error {
+      reject("CLEAR_USER_PERSONA", "Error removing User Persona", error)
+    }
+  }
+  
+  @objc(clearUserPersona:severity:properties:resolver:rejecter:)
```
jpmunz, line 313:
```suggestion
  @objc(logMessageWithSeverityAndProperties:severity:properties:resolver:rejecter:)
```

packages/core/ios/RNEmbrace/EmbraceManager.swift
```diff
@@ -20,565 +20,599 @@ private let EVENT_TIMESTAMP_KEY = "timeStampMs"
...
+  }
+  
+  @objc(startView:resolver:rejecter:)
+  func startView(_ viewName: String, resolver resolve: @escaping RCTPromiseResolveBlock, rejecter reject: @escaping RCTPromiseRejectBlock) {
+    let startTimeDate = Date()
     
+    let span = Embrace.client?.buildSpan(name: "emb-screen-view")
+      .setAttribute(key:"view.name", value:viewName)
+      .setAttribute(key:"emb.type", value:"ux.view")
+      .setStartTime(time: startTimeDate)
```
jpmunz, line 343: shouldn't be needed, it will default to the current time
```suggestion
```

packages/core/ios/RNEmbrace/EmbraceManager.swift
```diff
@@ -20,565 +20,599 @@ private let EVENT_TIMESTAMP_KEY = "timeStampMs"
...
-    func addBreadcrumb(_ event: String, resolver resolve: @escaping RCTPromiseResolveBlock, rejecter reject: @escaping RCTPromiseRejectBlock) {
-        //TODO Refactor SPAN
-        //      do {
-        //        try Embrace.client?.add(event: event)
-        //          resolve(true)
-        //      } catch {
-        //          reject(false)
-        //      }
+    if spanId.isEmpty {
+      reject("START_SPAN_ERROR", "Failed to start span", nil)
```
jpmunz, line 352:
```suggestion
      reject("START_VIEW_ERROR", "Failed to start view", nil)
```

packages/core/ios/RNEmbrace/EmbraceManager.swift
```diff
@@ -20,565 +20,599 @@ private let EVENT_TIMESTAMP_KEY = "timeStampMs"
...
+      resolve(spanId)
+    }
+  }
+  
+  @objc(endView:resolver:rejecter:)
+  func endView(_ spanId: String, resolver resolve: @escaping RCTPromiseResolveBlock, rejecter reject: @escaping RCTPromiseRejectBlock) {
+    let endTimeDate = Date()
+    let endTimeMS = NSNumber(value: endTimeDate.timeIntervalSince1970 * 1000)
+    
+    stopSpan(spanId, errorCodeString: "", endTimeMs: endTimeMS, resolver: resolve, rejecter: reject)
```
jpmunz, line 363: shouldn't need to pass in end time, should default to current time if given a 0
```suggestion
    stopSpan(spanId, errorCodeString: "", endTimeMs: 0.0, resolver: resolve, rejecter: reject)
```

jpmunz (approved):

PR #72, opened 2024-08-16
EMBR-4636 calculate react native bundle id

jpmunz:
On Android and previously on iOS 5.x we would take the given file path/url and compute the md5 hash of the contents and set this as `react_native_bundle_id` on the session. See:
* [Android](https://github.com/embrace-io/embrace-android-sdk/blob/master/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/capture/metadata/RnBundleIdTrackerImpl.kt)
* iOS
  * [EMBDevice.m](https://github.com/embrace-io/embrace-ios-sdk/blob/75415fae97c41ef50e8209b08addfa1aa51c6167/Embrace/EMBDevice.m#L577)
  * [EMBRNBundle.m](https://github.com/embrace-io/embrace-ios-sdk/blob/75415fae97c41ef50e8209b08addfa1aa51c6167/Embrace/EMBRNBundle.m#L22)
  * [NSData+Embrace.m](https://github.com/embrace-io/embrace-ios-sdk/blob/75415fae97c41ef50e8209b08addfa1aa51c6167/Embrace/NSData%2BEmbrace.m#L289)

This attempts to recreate that functionality in our RN Swift native module


packages/core/ios/RNEmbrace/BundleID.swift
```diff
@@ -0,0 +1,67 @@
...
+func computeBundleID(path: String) throws -> BundleID {
+    if path.isEmpty {
+        throw ComputeBundleIDErrors.emptyPath
+    }
+
+    var last = LastBundleComputation.fromStorage()
+
+    // Check if we can re-use our last computed bundle ID, this is true only if the path hasn't changed and
+    // the file contents at that path haven't been modified since the last time we computed the ID
+    // TODO in ios 5.x and Android implementation we are just returning last.id in this case without checking modified at...is that correct?
```
jpmunz, line 45: @ArielDemarco @TheMutsi this logic feels correct to me, e.g. only re-use the cache if the path is the same and the contents haven't been updated but it does deviate from the previous implementations so wanted to double-check

packages/core/ios/RNEmbrace/BundleID.swift
```diff
@@ -0,0 +1,67 @@
...
+        if let lastComputed = last.computedAt, let modifiedAt = attributes?[FileAttributeKey.modificationDate] as? Date,
+           modifiedAt <= lastComputed {
+            return BundleID(id: last.id, cached: true)
+        }
+    }
+
+    let fileContents = try String(contentsOf: URL(fileURLWithPath: path))
+
+    // https://stackoverflow.com/a/56578995
+    let bundleID = Insecure.MD5.hash(data: Data(fileContents.utf8)).map {
```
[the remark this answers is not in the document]
jpmunz: Added test to the upload tool to confirm it generates the same ID: https://github.com/embrace-io/go/pull/6958

facostaembrace (approved):

packages/core/ios/RNEmbrace/BundleID.swift
```diff
@@ -0,0 +1,67 @@
...
+
+    static func fromStorage() -> LastBundleComputation {
+        if let storedData = UserDefaults.standard.object(forKey: LastBundleComputation.STORAGE_KEY) as? Data,
+           let loaded = try? JSONDecoder().decode(LastBundleComputation.self, from: storedData) {
+            return loaded
+        }
+
+        return LastBundleComputation(path: "", computedAt: nil, id: "")
+    }
+}
```
ArielDemarco, line 25: I'd inject an "own" `UserDefaults` to prevent any problem in the future (like sombody migrating to another suite stuff, or things like that).
The only bad thing about injecting it (and having a instance variable with an own `userDefaults` is that you'll have to implement `Codable` manually.
So, i'd suggest something in the middle, like providing a utility to create an "Embrace" UserDefaults, like this:
```swift
extension UserDefaults {
    static var embrace: UserDefaults {
        UserDefaults(suiteName: "EmbraceReactNative") ?? .standard
    }
}
```
And then you can replace `store` and `fromStorage` using that utility:
```swift

func store () {
        if let encoded = try? JSONEncoder().encode(self) {
            UserDefaults.embrace.set(encoded, forKey: LastBundleComputation.STORAGE_KEY)
        }
    }

    static func fromStorage() -> LastBundleComputation {
        if let storedData = UserDefaults.embrace.object(forKey: LastBundleComputation.STORAGE_KEY) as? Data,
           let loaded = try? JSONDecoder().decode(LastBundleComputation.self, from: storedData) {
            return loaded
        }

        return LastBundleComputation(path: "", computedAt: nil, id: "")
    }
```

packages/core/ios/RNEmbrace/BundleID.swift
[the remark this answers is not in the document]
ArielDemarco: Yes; it simpler using `CryptoKit`. Main difference is that 5.x was supporting iOS 11 (as far as I remember), so this was unavailable, and most APIs were C-alike

packages/core/test-project/ios/RNEmbraceTests/RNEmbraceTests.swift
```diff
@@ -799,3 +799,70 @@ class EmbraceSpansSDKNotStartedTests: XCTestCase {
     }
 }
 
+class ComputeBundleIDTests: XCTestCase {
+    // https://nshipster.com/temporary-files/
```
ArielDemarco, line 803: Probably many of this tests under the hood, are using `UserDefaults`. `UserDefaults` writes things to disk.
So, to prevent any kind of flakyness, i'd remove all things persisted in UserDefaults between tests. In order to achieve this (considering you've [made this change](https://github.com/embrace-io/embrace-react-native-sdk/pull/72/files#r1731339931), i'd add this code:

```swift
    override func tearDownWithError() throws {
          UserDefaults.standard.removePersistentDomain(forName: "EmbraceReactNative")
    }
```

ArielDemarco (approved):

packages/core/ios/RNEmbrace/BundleID.swift
ArielDemarco: LGTM; I'd wait for @TheMutsi to see if there's nothing we're not considering

packages/core/ios/RNEmbrace/BundleID.swift
jpmunz: makes sense, added

packages/core/test-project/ios/RNEmbraceTests/RNEmbraceTests.swift
jpmunz: added

packages/core/ios/RNEmbrace/BundleID.swift
TheMutsi: @ArielDemarco can we move it to DispatchQueue ? I think it could be very heavy

packages/core/ios/RNEmbrace/EmbraceManager.swift
```diff
@@ -204,23 +204,22 @@ class EmbraceManager: NSObject {
...
-        //             resolve(false)
-        // #endif
-        //         } catch let error {
-        //             reject("CHECK_AND_SET_CODEPUSH_BUNDLE_URL", "Error setting Codepush Bundle URL", error)
-        //         }
+                 do {
+         #if canImport(CodePush)
+                     guard let url = try CodePush.bundleURL() else {
+                         reject("GET_CODEPUSH_BUNDLE_URL", "Error getting Codepush Bundle URL", nil)
+                         return
```
TheMutsi, line 211: I wonder if it's better not maintain "codepush" in a specific method, and just let the user set the javascript url if they are using codepush. @jpmunz  what do you think? we can just remove this function and update the doc so we dont care if Codepush changes its name or method in the future

jpmunz: I think I'd keep it as is for the purposes of the iOS 6.x migration but I'd like to revisit along with calculating the bundle ID altogether to see if we can reduce code here, noted in https://www.notion.so/embraceio/Rethink-maintaining-support-for-OTA-updates-efe0b4b089ba487d9ab6678b7f7ca491?pvs=4

packages/core/ios/RNEmbrace/BundleID.swift
jpmunz: I think that makes sense, I moved the calls to compute ID to background queues

jpmunz:
@ArielDemarco @facostaembrace @TheMutsi a bit of a larger change in https://github.com/embrace-io/embrace-react-native-sdk/pull/72/commits/cf514f51a27d3480eb39b7d273caef204f94a939 to get this working if you wanted to recheck the PR

previously when the RN ios module was written in objective-c we had wrapped the code push in a `#if __has_include(<CodePush/CodePush.h>)` directive to only run if the host app actually had code push. The equivalent `#if canImport(CodePush)` for Swift didn't work as code push isn't exposed as a Swift module.

Instead I ended up adding a new objective-c `CodePushHelper` class which all it does is wrap the call we need to grab the codepush url in that same directive we previously had or nil otherwise, then on the swift side we call into that without any compiler directives

PR #75, opened 2024-08-22
Rn 4.2.1

TheMutsi:
## Goal

Update the Android SDK to include the refactor for source map and bundle uploads. This update also includes using MD5 instead of the native build ID.

Extra: I added some files to the package.json to exclude some files from the final package

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -15,8 +15,14 @@
   },
   "files": [
     "lib",
-    "android",
-    "ios",
+    "android/src",
+    "android/build.gradle",
+    "android/gradle.properties",
+    "ios/RNEmbrace/EmbraceManager.h",
+    "ios/RNEmbrace/EmbraceManager.m",
+    "ios/RNEmbrace/EmbraceOTelSpanHelper.h",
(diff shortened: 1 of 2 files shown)


facostaembrace (approved):

packages/core/package.json
```diff
@@ -15,8 +15,14 @@
   },
   "files": [
     "lib",
-    "android",
-    "ios",
+    "android/src",
+    "android/build.gradle",
+    "android/gradle.properties",
+    "ios/RNEmbrace/EmbraceManager.h",
```
jpmunz, line 21: does anything unwanted get added if we use `ios/RNEmbrace` here instead? I'm just a bit worried about forgetting to update this when add/remove files like we did on the 5.0 branch

TheMutsi: Yes, the info.plist for now.

TheMutsi:
@facostaembrace  I will close this PR since @jpmunz asked to bump to 4.3.0 since android requiere an agp/gradle bump version. I will delete this branch  and open a new one

PR #83, opened 2024-08-24
[EMBR-4352] Run Swiftlint/Clang-format over RN iOS module

facostaembrace:
## Test

> given this [failure](https://github.com/embrace-io/embrace-react-native-sdk/actions/runs/10534470261/job/29192402283) had to revert this format [change](https://github.com/embrace-io/embrace-react-native-sdk/pull/83/commits/5b7922171b74ddbbd0a135f220ddf48fd023d683). To analyze how this string is working and if it is safe just to fix the test with the new format of the file.

UPDATE: previous issue was fixed. Unit tests work fine after tweaks on strings.

## Goal

- adding Swiftlint for swift files (`yarn lint:swift`)
- adding linter for Objective-c files (`yarn lint:clang`)
- adding script for linting js files only into packages/** directory (`yarn lint:js`). NOTE: the base `yarn lint` cover json files, and json files are also into ios/android directories making the plugin to run into the entire project (resulting in unexpected behaviors)
- adding husky to run linters when commiting


package.json
```diff
@@ -10,15 +10,24 @@
     "node": ">=18.0.0"
   },
   "scripts": {
+    "prepare": "husky",
     "publish-modules": "npx lerna run build && npx lerna publish",
     "build": "npx lerna run build",
+    "test": "jest",
     "lint": "eslint . --ext .js,.jsx,.ts,.tsx --fix && prettier --write \"**/*.{js,jsx,ts,tsx,json}\" && yarn constraints --fix",
```
jpmunz, line 17: is this regular `lint` command still needed given the others we have now?

packages/core/ios/RNEmbrace/EmbraceManager.m
jpmunz, line 1: this one will conflict since we fully changed it on 5.0, might be better to not update as part of this PR

packages/core/scripts/util/ios.ts
```diff
@@ -18,8 +18,8 @@ export const exportSourcemapRNVariable =
   'export SOURCEMAP_FILE="$CONFIGURATION_BUILD_DIR/main.jsbundle.map";';
 
 export const EMBRACE_IMPORT_OBJECTIVEC = "#import <Embrace/Embrace.h>";
-export const EMBRACE_INIT_OBJECTIVEC =
-  "[[Embrace sharedInstance] startWithLaunchOptions:launchOptions framework:EMBAppFrameworkReactNative];";
+export const EMBRACE_INIT_OBJECTIVEC = `[[Embrace sharedInstance] startWithLaunchOptions:launchOptions
+                                           framework:EMBAppFrameworkReactNative];`;
```
jpmunz, line 22: I'm not sure about this change, since it's injected into the customer's app delegate I think this will end up including extra whitespace

jpmunz (approved):
lgtm, I think this would be good to hook up to the github workflow as well

when merging 5.0 we should remember remove the extra `.swiftlint.yml` files added there

packages/core/ios/RNEmbrace/EmbraceManager.m
facostaembrace: Yes, this is why I would prefer to wait until we make the releases and then after all merge this pr

package.json
jpmunz: ya I think we can just remove

PR #84, opened 2024-08-26
[EMBR-4464] Hook Swift tests up to CI

facostaembrace:
## Goal

- Adding a workflow that handles iOS Unit Test.
- TBD to add a labels and decide whether the iOS unit tests should run


packages/core/test-project/ios/RNEmbraceTestProject.xcodeproj/xcshareddata/xcschemes/RNEmbrace.xcscheme
facostaembrace, line 1: NOTE: to cleanup this scheme, this is not needed (RNEmbraceTests is the one needed here)

.github/workflows/ios.unit-test.yml
```diff
@@ -0,0 +1,71 @@
...
+      run: |
+        cd ${{ env.PROJECT_PATH }}
+        yarn install
+        cd ios
+        pod install
+    
+    - name: Run iOS Unit Tests
+      run: |
+        cd ${{ env.PROJECT_PATH }}/ios
+        xcodebuild test \
```
jpmunz, line 59: could this just call `yarn ios:test` ?

packages/core/test-project/ios/.gitignore
```diff
@@ -16,11 +16,9 @@ build/
 xcuserdata
 *.xccheckout
 *.moved-aside
-DerivedData
```
jpmunz, line 19: is it necessary to remove `DerivedData` here? I believe these are re-generated on each local build so could create noise in git

PR #158, opened 2024-09-27
EMBR-4229 integration tests for using the otel navigation library

jpmunz:
Adding integration tests for using the otel navigation instrumentation library with our tracer provider.

Follow-ups outside of this PR:
* There seems to be a type error when passing `ref` to the NavigationTracker exposed by the library: https://www.notion.so/embraceio/Type-error-when-passing-ref-to-instrumentation-react-native-navigation-NavigationTracker-10e7e3c99852805cbaa4deb81ef195c5?pvs=4
* We may want to make this setup easier on customers when we switch to this from our Embrace navigation package in 6.x: https://www.notion.so/embraceio/Add-helpers-or-a-bundled-package-that-makes-integrating-the-different-sdk-features-plug-and-play-10c7e3c9985280f3b1e0fa0b28b5a570?pvs=4


packages/react-native-tracer-provider/src/EmbraceNativeTracerProvider.ts
```diff
@@ -56,13 +61,13 @@ export const useEmbraceNativeTracerProvider = (
...
         .then((started: boolean) => {
-          if (started) {
-            setTracerProvider(new EmbraceNativeTracerProvider(config));
-          } else {
+          if (!started) {
             setError(
               "The Embrace SDK must be started to use the TracerProvider, please invoke `initialize` from `@embrace-io/react-native`.",
             );
             setIsError(true);
+          } else if (!tracerProvider) {
```
jpmunz, line 69: @facostaembrace I couldn't reproduce seeing the view span recorded twice after making this change so I'll assume for now it was due to accidentally instantiating this provider multiple times

integration-tests/specs/navigation.test.ts
```diff
@@ -0,0 +1,154 @@
...
+import {getSpanPayloads} from "../helpers/embrace_server";
+import {EmbraceSpanData} from "../typings/embrace";
+import {
+  commonEmbraceSpanAttributes,
+  embraceSpanDefaults,
+  sortSpanAttributes,
+} from "../helpers/span";
+
+describe("Navigation", () => {
+  it("should record the initially rendered screen", async () => {
```
jpmunz, line 13: @facostaembrace I was seeing this passing on both Android and iOS which I believe addresses the concern of not capturing the first render, unless there is another test you think could be added?

integration-tests/basic-test-app/app/_layout.tsx
```diff
@@ -1,25 +1,31 @@
...
+  const {tracerProvider} = useEmbraceNativeTracerProvider({}, embraceSDKLoaded);
+  const navigationRef = useNavigationContainerRef();
+
   useEffect(() => {
     const init = async () => {
       const hasStarted = await initEmbrace({
         sdkConfig: {
           ios: {
-            appId: "abcdf",
+            appId: "cvKeD",
```
facostaembrace, line 26: do we want to commit the real app id of your test app?

integration-tests/specs/navigation.test.ts
facostaembrace: I think it's just fine. We can add more test if we see other weirdness happening

facostaembrace (approved):

integration-tests/basic-test-app/app/_layout.tsx
jpmunz: reverting for now, we may want to revisit this as part of integration testing as to what we want this to default to

PR #169, opened 2024-10-03
EMBR-5028 only run ios test workflow when PR contains ios changes

jpmunz:


.github/workflows/ios.unit-test.yml
```diff
@@ -33,19 +33,33 @@ jobs:
...
+            - 'packages/**/*.h'
+
       - name: Cache Derived Data
         uses: actions/cache@v3
         with:
           path: ~/Library/Developer/Xcode/DerivedData
           key: Derived-Data-${{ runner.os }}-${{ hashFiles('**/*.xcodeproj/project.pbxproj, Podfile.lock') }}
           restore-keys: |
             Derived-Data-${{ runner.os }}
+        if: steps.changed_files.outputs.ios == 'true'
```
jpmunz, line 53: @embrace-io/team-devops was curious to get your opinion here, we're wanting to not run the iOS unit tests if we don't detect any changes to iOS source files. It seemed odd to have to make this check on every step but googling around briefly I couldn't see a better way, do we do something similar in any other workflow?

vitaliyf: Doing what you did is pretty reasonable. It's somewhat similar to what we do in "go"

Otherwise, there is https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#running-your-pull_request-workflow-based-on-files-changed-in-a-pull-request but this would make entire workflow run or not run based on paths. You would then need to split into separate "workflows/ios" vs "workflows/not-ios" jobs for each set of filename patterns. There isn't anything wrong with doing that, other than you could end up copy/pasting much of it between them (which could then be mitigated by reusing a workflow..)

jpmunz: got it

ya I think splitting of the workflow could make sense, the earlier steps are common to the Android and JS testing workflows as well so that could be a reusable workflow. I'll file that as follow-up tech debt if this seems reasonable I don't think that would be blocking and more a nice to have

facostaembrace (approved):

PR #194, opened 2024-10-15
[EMBR-4352] Run Swiftlint/Clang-format over RN iOS module

facostaembrace:
## Test

> given this [failure](https://github.com/embrace-io/embrace-react-native-sdk/actions/runs/10534470261/job/29192402283) had to revert this format [change](https://github.com/embrace-io/embrace-react-native-sdk/pull/83/commits/5b7922171b74ddbbd0a135f220ddf48fd023d683). To analyze how this string is working and if it is safe just to fix the test with the new format of the file.

UPDATE: previous issue was fixed. Unit tests work fine after tweaks on strings.

## Goal

- adding Swiftlint for swift files (`yarn lint:swift`)
- adding linter for Objective-c files (`yarn lint:clang`)
- adding script for linting js files only into packages/** directory (`yarn lint:js`). NOTE: the base `yarn lint` cover json files, and json files are also into ios/android directories making the plugin to run into the entire project (resulting in unexpected behaviors)
- adding husky to run linters when commiting

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,17 +10,26 @@
     "node": ">=18.0.0"
   },
   "scripts": {
+    "prepare": "husky",
     "publish-modules": "npx lerna run build && npx lerna publish",
     "build": "npx lerna run build",
     "lint": "eslint . --ext .js,.jsx,.ts,.tsx --fix && prettier --write \"**/*.{js,jsx,ts,tsx,json}\" && yarn constraints --fix",
+    "lint:js": "eslint . --ext .js,.jsx,.ts,.tsx --fix && prettier --write \"packages/**/*.{js,jsx,ts,tsx,json}\" && yarn constraints --fix",
+    "lint:clang": "bash ./scripts/run-clang-format.sh",
+    "lint:swift": "bash ./scripts/run-swiftlint.sh",
     "test": "jest",
     "ios:install": "npx lerna run ios:install",
     "ios:test": "npx lerna run ios:test"
   },
   "lint-staged": {
-    "**/*.{js,jsx,ts,tsx}": [
-      "yarn lint"
-    ]
+    "*.{js,jsx,ts,tsx}": "yarn lint:js",
+    "*.{m,mm,h}": "yarn lint:clang",
+    "*.swift": "yarn lint:swift"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
   },
   "devDependencies": {
     "@babel/core": "^7.9.0",
@@ -44,6 +53,7 @@
     "eslint-plugin-prettier": "^5.1.3",
     "eslint-plugin-react": "^7.34.1",
     "eslint-plugin-react-native": "^4.1.0",
+    "husky": "^9.1.5",
     "jest": "^29.7.0",
     "lerna": "^8.1.5",
     "lint-staged": "^15.2.10",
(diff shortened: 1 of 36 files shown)


scripts/run-clang-format.sh
```diff
@@ -0,0 +1,72 @@
...
+FIND_CMD="find"
+for dir in "${SOURCE_DIRS[@]}"; do
+    FIND_CMD+=" $dir"
+done
+
+# 2) Exclude directories. Add more directories if needed
+EXCLUDE_DIRS=(
+    "$(pwd)/node_modules"
+    "$(pwd)/build"
+    "$(pwd)/packages/core/scripts/__tests__"
```
facostaembrace, line 32: @jpmunz added this line here to avoid linting objective-c files under __tests__ since we are testing and comparing exact strings (which ends into failing cases)

package.json
```diff
@@ -10,17 +10,25 @@
     "node": ">=18.0.0"
   },
   "scripts": {
+    "prepare": "husky",
     "publish-modules": "npx lerna run build && npx lerna publish",
     "build": "npx lerna run build",
-    "lint": "eslint . --ext .js,.jsx,.ts,.tsx --fix && prettier --write \"**/*.{js,jsx,ts,tsx,json}\" && yarn constraints --fix",
+    "lint:js": "eslint . --ext .js,.jsx,.ts,.tsx --fix && prettier --write \"packages/**/*.{js,jsx,ts,tsx,json}\" && yarn constraints --fix",
+    "lint:clang": "bash ./scripts/run-clang-format.sh",
```
jpmunz, line 17: minor nit but for consistency with scripts like `run-ios-tests.sh` you can remove bash here and add `#!/bin/bash` as the first line in the script, then can just have `./scripts/run-clang-format.sh`here

packages/core/ios/RNEmbrace/EmbraceManager.m
```diff
@@ -1,181 +1,137 @@
...
+RCT_EXTERN_METHOD(addSpanEventToSpan : (NSString *)spanId name : (NSString *)name time : (double)
+                      time attributes : (NSDictionary)attributes resolver : (RCTPromiseResolveBlock)
+                          resolve rejecter : (RCTPromiseRejectBlock)reject)
+
+RCT_EXTERN_METHOD(addSpanAttributeToSpan : (NSString *)spanId key : (NSString *)
+                      key value : (NSString *)value resolver : (RCTPromiseResolveBlock)
+                          resolve rejecter : (RCTPromiseRejectBlock)reject)
+
+RCT_EXTERN_METHOD(
+    recordCompletedSpan : (NSString *)name startTimeMs : (double)startTimeMs endTimeMs : (double)
```
jpmunz, line 100: minor but I kind of find the previous version a bit more readable here, are there any quick configs around this we could tweak? In particular I feel we don't need the extra spaces in ` recordCompletedSpan : (NSString *)name` as well as the extra newlines/tabs

packages/core/scripts/__tests__/__mocks__/android/buildWithSwazzler.gradle
jpmunz, line 1: can be reverted, may need to revisit this I thought it had been fixed but I believe these lines get added on every run of the tests

scripts/run-swiftlint.sh
jpmunz, line 1: makes sense to introduce the `scripts` folder, once this is merged we can move `run-ios-tests.sh` and `scopes-from-file-changes.ts` into it as well

jpmunz (approved):

packages/core/ios/RNEmbrace/EmbraceManager.m
facostaembrace: just reverted the changes for this file and disabled formatter for .m by now, I can revisit it later on

PR #199, opened 2024-10-16
Get data from mock api

TheMutsi:
## Release Notes

**WHAT**:<br>
**WHY**:<br>
**WHO**:<br>


integration-tests/specs/session.test.ts
```diff
@@ -1,17 +1,3 @@
-import {driver} from "@wdio/globals";
-
-import {getSessionPayloads} from "../helpers/embrace_server";
-
 describe("Sessions", () => {
-  it("should be recorded as foreground", async () => {
```
jpmunz, line 6: why was this removed?

integration-tests/typings/embrace.d.ts
```diff
@@ -14,4 +14,17 @@ interface EmbraceRequestBody {
   json: EmbracePayload;
 }
 
-export type {EmbraceRequestBody, SessionPayload};
+interface Data {
```
jpmunz, line 17: will be looking to merge https://github.com/embrace-io/embrace-react-native-sdk/pull/53/files#diff-3c4284e57c0d8ce1dae6e29d723e3535d91d54939e07856eee2b47de6c02d872 shortly which has some changes to the helpers and typings here that we may need to resolve

integration-tests/helpers/embrace_server.ts
```diff
@@ -1,44 +1,52 @@
...
-
-const clearServer = async () => {
-  return await mockServerClient("localhost", PORT).clear({}, "LOG");
-};
-
-const getSessionPayloads = async (delay = 2000): Promise<SessionPayload[]> => {
+import {EmbraceData} from "../typings/embrace";
+import axios from "axios";
+
+// TODO Change it to dot env or some variable
```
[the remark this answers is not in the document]
TheMutsi: Sure

integration-tests/specs/session.test.ts
TheMutsi: Because this is only to get data from the API, in the other PR that is based on this one. has all the test

integration-tests/package.json
```diff
@@ -19,8 +19,10 @@
     "appium": "^2.6.0",
     "appium-uiautomator2-driver": "^3.5.2",
     "appium-xcuitest-driver": "^7.17.2",
+    "axios": "^1.7.7",
```
jpmunz, line 22: are both axios and node-fetch needed?

integration-tests/specs/session.test.ts
jpmunz: until we have replacements I think reasonable to leave this one here as quick example

integration-tests/typings/embrace.d.ts
```diff
@@ -14,4 +14,17 @@ interface EmbraceRequestBody {
...
+  Body: string;
+}
+interface EmbraceData {
+  Spans: IData[];
```
jpmunz, line 21: we should be good with just `Spans` and `Logs` here, the rest are part of the v1/ endpoints and no longer used

integration-tests/.env
```diff
@@ -0,0 +1,4 @@
+RAW_EMB_URL=
```
jpmunz, line 1: can you update `integration-tests/README.md` with an example of how the integration tests should now be run with these env vars? The old instructions could be removed as well as `helpers/invoke_embrace_server.ts`

integration-tests/helpers/embrace_server.ts
```diff
@@ -1,44 +1,50 @@
...
-  });
-};
-
-const clearServer = async () => {
-  return await mockServerClient("localhost", PORT).clear({}, "LOG");
-};
-
-const getSessionPayloads = async (delay = 2000): Promise<SessionPayload[]> => {
+const getSessionPayloads = async (
+  sessionId: string | string[],
```
jpmunz, line 8: instead of the caller needing to figure out the session ID I think simpler if in `beforeTest` a DELETE request was made to the namespace to clear out any data generated by previous testing, then we could grab all session payloads here

integration-tests/typings/embrace.d.ts
TheMutsi: ok

integration-tests/helpers/embrace_server.ts
TheMutsi: If you do that, you won't be able to run multiple instances using the same test APK. You will need to create different APKs with different embrace config

jpmunz: hmm that's true, ok let's keep for now we may want to revisit when we hook things up to CI depending on how much we run concurrently

integration-tests/typings/embrace.d.ts
```diff
@@ -60,8 +60,24 @@ interface EmbracePayload {
...
 
+interface EmbraceRequestBody {
+  json: EmbracePayload;
+}
+
+interface Data {
+  Body: string;
+}
+interface EmbraceData {
+  Spans: Data[];
```
jpmunz, line 71: for spans should use the existing `ParsedSpanPayload` type

integration-tests/helpers/embrace_api.ts
```diff
@@ -0,0 +1,46 @@
...
+    });
+
+  const dataFiltered = {
+    Spans: [],
+    Logs: [],
+  };
+
+  Object.entries(responseData).forEach(([key, value]) => {
+    dataFiltered[key] = getDataFromSessionId(value, sessionId).map(item =>
+      JSON.parse(item.Body),
```
jpmunz, line 39: for spans should parse this using `parseSpanPayload` that's used in `helpers/embrace_server.ts`. Actually should get rid of `helpers/embrace_server.ts` altogether or have this replace it, ideally the specs that are calling it would behave the same if they get back the same ParsedSpanPayload type that was being used before from the local mocked server

integration-tests/helpers/session.ts
```diff
@@ -1,5 +1,37 @@
 import {driver} from "@wdio/globals";
 
+const SHOW_SESSION_ID_TEXT_BUTTON = "~SHOW SESSION ID";
```
jpmunz, line 3: the test harness will also need to be updated to make sure this button is shown on each screen and each of the existing test specs to use it before grabbing the span payloads

thinking more on this I do think it's simpler to instead DELETE the content in the namespace on the mock-server before each test. The downside of not being able to re-use the same app package concurrently in different tests runs I think is maybe less of a hassle in the long run, we'd need to make sure as part of CI/CD that we generate one APK per OS version we want to test and give them each a different namespace

integration-tests/helpers/span_util.ts
jpmunz, line 1: should revisit these helpers after using `parseSpanPayload` and moving any that are still needed into `integration-tests/helpers/span.ts`

integration-tests/specs/session.test.ts
jpmunz, line 1: for the assertions here should see if a style closer to `integration-tests/specs/navigation.test.ts` and `integration-tests/specs/tracer_provider.test.ts` could work where the spans are compared against an expected JSON array and we plugin default values or filter things out that we don't need in the comparison

integration-tests/helpers/session.ts
TheMutsi: We can put that button on the app header or footer, so it will be there in any Page.

jpmunz:
replaced by https://github.com/embrace-io/embrace-react-native-sdk/pull/238

PR #212, opened 2024-10-22
EMBR-5083 -- Template for creating an Expo test app

jpmunz:
Added an expo template for creating a test app along with a new helper script for changing the embrace configuration for a particular test app, using these I was able to do:

```
npx create-expo --template ./templates/expo-test-app-template/expo-test-app-template-1.0.0.tgz
> foobar-app

./update-local-embrace.sh foobar-app
./set-embrace-config.js foobar-app ./embrace-configs/local-node-server.json

cd foobar-app
npx expo run:android / run:ios
```

and then run the integration tests on the new app

Not committing the new app here, just the template files for it, will revisit when looking at testing across RN versions whether we want to commit more example apps or just generate them on the fly

Next steps:
* Similar template for creating a bare react native app
* Workflow/tooling for integration testing across all React Native versions we want to support


integration-tests/templates/expo-test-app-template/android/app/build.gradle
jpmunz, line 1: the template files unfortunately make the diff pretty large, it's mostly boilerplate and was copied from `basic-test-app/`

integration-tests/templates/expo-test-app-template/ios/HelloWorld.xcodeproj/project.pbxproj
jpmunz, line 1: instances of "HelloWorld" get replaced by the package name passed to `create-expo`

integration-tests/update-local-embrace.sh
```diff
@@ -40,14 +58,17 @@ if [ "$test_harness_only" = false ]; then
...
-  ./artifacts/opentelemetry-instrumentation-react-native-navigation-0.1.0.tgz \
-  ./artifacts/embrace-io-react-native-test-harness-local.tgz
\ No newline at end of file
+if [ "$skip_test_harness" = false ]; then
+  # build and pack the test harness
+  npm --prefix test-harness run build
+  ./pack.sh  test-harness/ artifacts/embrace-io-react-native-test-harness-local.tgz
+fi
+
+if [ "$skip_sdk_packages" = false ] || [  "$skip_test_harness" = false ]; then
```
facostaembrace, line 67: let me know if I'm wrong but if I pass --skip-sdk-packages (so `skip_sdk_packages` is true) and do not pass --skip-test-harness (`skip_test-harness` is false) then the script will still try to install everything, right?

but since `skip_sdk_packages` is `true` the sdk packages won't be packed, so there will not be there. will this throw an error? (same thing would apply if I pass --skip-test-harness and don't pass --skip-sdk-packages)

integration-tests/update-local-embrace.sh
```diff
@@ -40,14 +58,17 @@ if [ "$test_harness_only" = false ]; then
...
+  # update the test app with the sdk packages and the test harness
+  npm --prefix $test_app add \
+    ./artifacts/embrace-io-react-native-local.tgz \
+    ./artifacts/embrace-io-react-native-tracer-provider-local.tgz \
+    ./artifacts/opentelemetry-instrumentation-react-native-navigation-0.1.0.tgz \
```
facostaembrace, line 72: I don't see `opentelemetry-instrumentation-react-native-navigation-0.1.0` packed under the if block of `skip_sdk_packages`, is this meant to be packed manually? (if so, it makes sense because it's coming from the other repo by now, but just trying to confirm)

facostaembrace: my bad! `opentelemetry-instrumentation-react-native-navigation-0.1.0.tgz` is actually committed in `integration-tests/artifacts`. ignore this comment

integration-tests/templates/expo-test-app-template/android/app/src/main/res/mipmap-hdpi/ic_launcher.png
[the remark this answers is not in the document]
jpmunz: let me take a look, unlike the ones under /native-src I think some of these may be needed since we're actually launching the app but maybe there's a default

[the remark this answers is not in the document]
jpmunz: was getting some errors on iOS related to the splashscreen so kept `Images.xcassets/` but was able to remove most of the other images as well as a bunch of extra stuff at the JS level that isn't needed, delete them from basic-test-app as well

integration-tests/update-local-embrace.sh
jpmunz: updated

integration-tests/update-local-embrace.sh
jpmunz: added a quick note about it to the script

facostaembrace (approved):

PR #215, opened 2024-10-24
Jpmunz/embr 5373 allow testing all supported rn versions

jpmunz:
The original plan was to have a smaller number of templates in `integration-tests/templates/` that could each support a range of React Native versions and then introduce a helper script that would be able to automatically switch the RN version used in order to test a different one

In testing this out I realized that the differences between minor versions are large enough that the script would have to be fairly complicated. Wanted to explore this alternate approach where there would be a separate template for each minor version we want to test as well as a smaller number that use expo, leading to something like:
```
integration-tests/templates/
    rn74-template/
    rn73-template/
    rn72-template/
    rn71-template/
    rn70-template/
    expo-rn74-template/
```

This simplifies wanting to test a specific RN version as you would then just create an app from the associated template. The downside is a lot of files introduced to the repo as well as a fair amount of duplication between each template as most files remain the same


packages/react-native-tracer-provider/native-src/android/app/src/main/java/io/embrace/reactnativetracerprovider/ReactNativeTracerProviderModule.kt
jpmunz, line 1: Separate bug fix, realized in testing RN 0.73 that this newer syntax wasn't compiling on older versions

integration-tests/templates/react-native-test-app-template/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "react-native-test-app-template",
+  "name": "rn74-template",
```
facostaembrace, line 2: shoult the name of the directory be `rn74-template` as per this name?

facostaembrace (approved):

integration-tests/templates/react-native-test-app-template/package.json
jpmunz: updated

PR #233, opened 2024-10-28
EMBR-4920 integration testing on Browserstack

jpmunz:
Adding a workflow for running a simple test (just make sure Embrace SDK has loaded by clicking a button) across apps using multiple RN versions and across a couple different OS versions for Android + iOS. The plan would be to run these against any `release/` branch we create as part of QA. It takes about an hour to finish all the combinations which isn't too bad if we're only doing it for a release but if it becomes annoying we can pay more for our browserstack plan to get additional parallel test runs to speed things up.

PR is a bit larger than I had originally planned, I ended up doing cleanup across the existing integration test tooling+templates to get things working with browserstack. Things are generally succeeding now though there appears to still be some flakiness with the react native template building giving "Couldn't find the "...template.config.js file" errors that I will look into.


integration-tests/update-embrace-packages.sh
jpmunz, line 1: renamed from `integration-tests/update-local-embrace.sh`

packages/react-native-otlp/package.json
```diff
@@ -31,5 +31,29 @@
   "embrace": {
     "iosVersion": "6.5.1",
     "androidVersion": "6.13.0"
-  }
+  },
+  "author": "Embrace <support@embrace.io> (https://embrace.io/)",
```
jpmunz, line 35: building on iOS complains if these are missing

jpmunz: although with it in place the linter fails since it isn't expected these attributes for a private package

integration-tests/templates/rn71-template/App.tsx
```diff
@@ -7,7 +7,7 @@ function App(): React.JSX.Element {
     <EmbraceTestHarness
       navigationStyle="react-native"
       sdkConfig={sdkConfig}
-      allowCustomExport={true}
+      allowCustomExport={false}
```
jpmunz, line 10: @facostaembrace I originally switched this to false for testing because the package wasn't available for Android, however even though that's available now I'm thinking it may make sense to keep it set as false since that will be the more common code path our users take, what do you think?

facostaembrace: I think I agree on that, probably I can try to run this suite locally following the steps you mentioned earlier to confirm that the new otlp works fine for all rn versions before I release the component. that should be enough for testing purposes. but I agree that we don't need this turned on for every release

packages/react-native-otlp/package.json
facostaembrace: will remove the private key soon. I'll keep you posted

facostaembrace (approved):

integration-tests/templates/rn71-template/App.tsx
jpmunz: another option maybe would be to commit an app to the repo that is configured for custom export and then have certain tests that only run with that one specifically

or actually maybe simpler would be to add some button to the test harness that restarts the app and flips it to use custom export

DEVELOPING.md
```diff
@@ -58,14 +58,15 @@ unreleased changes on `main` and a patch release will be cut from that new branc
...
 
-1. Create a release branch off of main and push it to origin
-2. Make sure you are logged into the npmjs registry (`npm login`)
-3. Release to npm with `yarn publish-modules`, you will be prompted to choose the version number to update to
-4. Check https://www.npmjs.com/org/embrace-io, the latest versions should have been published
-5. Check https://github.com/embrace-io/embrace-react-native-sdk/tags, a vX.X.X tag should have been pushed
-6. Create a PR from your release branch against `main` to merge all the version updates
-7. Run an example app and point to the latest released packages to confirm basic behaviour
-8. Update and publish the [Changelog](https://github.com/embrace-io/embrace-docs/blob/main/docs/react-native/changelog.md) for the release
+1. Create a `release/` branch off of main and push it to origin
```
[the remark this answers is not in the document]
jpmunz: I updated the readme let me know if it makes more sense now. Also realized it's possible to open a PR before any changes are made on the release branch as long as we include an empty commit

PR #238, opened 2024-11-01
Add Capabilities, and logs, action buttons

TheMutsi:
## Goal

Adding mock api and simple log.


jpmunz (commented):
@TheMutsi can you clarify how this relates to https://github.com/embrace-io/embrace-react-native-sdk/pull/199 since both are switching to the mock-api? Should that other one be closed?

integration-tests/typings/embrace.d.ts
```diff
@@ -27,17 +27,64 @@ interface EmbraceSpanData {
   attributes: EmbraceSpanAttribute[];
 }
 
+interface EmbraceLogData {
```
jpmunz, line 30: should remove the span specific stuff here, e.g. `trace_id`, `status`, `events`. For attributes let's rename `EmbraceSpanAttribute` -> `EmbraceAttribute` since it's used for both

integration-tests/typings/embrace.d.ts
```diff
@@ -27,17 +27,64 @@ interface EmbraceSpanData {
...
+}
+
+interface ParsedLogPayload {
+  resource?: EmbracePayloadResource;
+  metadata?: EmbracePayloadMetadata;
+  info: EmbraceLogData[];
+  warn: EmbraceLogData[];
+  error: EmbraceLogData[];
+}
+interface EmbracePayloadLogs {
```
jpmunz, line 48: should follow the pattern for spans, this could just have `logs: EmbraceLogData[]` as a property and then below update `EmbracePayload` to have type: "spans" | "logs" and data: EmbracePayloadSpans | EmbracePayloadLogs

integration-tests/typings/embrace.d.ts
```diff
@@ -27,17 +27,64 @@ interface EmbraceSpanData {
...
+  severity_text: string;
+  body: string;
+  time_unix_nano: number;
+  status: "Unset" | "Ok" | "Error";
+  events: EmbraceSpanEvent[];
+  attributes: EmbraceSpanAttribute[];
+}
+
+interface ParsedLogPayload {
+  resource?: EmbracePayloadResource;
```
jpmunz, line 42: I don't think we need resource + metadata here if it's on the EmbracePayload

integration-tests/helpers/action.ts
```diff
@@ -0,0 +1,40 @@
+import {driver} from "@wdio/globals";
+interface ActionButtons {
```
jpmunz, line 2: I'm not sure the overhead of this abstraction is worthwhile, I think it'd be enough to have tests call `pressButton("foo")` directly

yarn.lock
jpmunz, line 1: delete here was accidental?

integration-tests/helpers/embrace_mock_api.ts
```diff
@@ -0,0 +1,37 @@
+import {ParsedLogPayload, ParsedSpanPayload} from "../typings/embrace";
+import axios from "axios";
+import {parseSpanPayload} from "./span";
+import {parseLogPayload} from "./log";
+
+const EMB_URL = "https://mock-api.emb-eng.com/namespace/SgNw5/stored";
```
jpmunz, line 6: namespace needs to be dynamic here

integration-tests/helpers/log.ts
```diff
@@ -0,0 +1,50 @@
...
+    resource: undefined,
+  };
+
+  payload.data.logs.forEach(log => {
+    const logType = EMBRACE_INTERNAL_LOG_SEVERITY_NUMBER[log.severity_number];
+    parsedLog[logType].push(log);
+  });
+  parsedLog.metadata = payload.metadata;
+  //TODO See how can we check these two attributes using REGEX or something
+  delete payload.resource.process_start_time;
```
jpmunz, line 39: I think we can just remove them from the type definition if we don't need them in the tests

integration-tests/specs/log.test.ts
```diff
@@ -0,0 +1,22 @@
...
+    await ACTION_BUTTONS.LOG.ERROR();
+    await ACTION_BUTTONS.LOG.WARN();
+    await new Promise(r => setTimeout(r, 1000));
+    await endSession();
+
+    const logPayloads = await getLogPayloads();
+    expect(logPayloads).toHaveLength(1);
+    const [logPayload] = logPayloads;
+
+    Object.values(EMBRACE_INTERNAL_LOG_SEVERITY_NUMBER).forEach(logSeverity =>
```
jpmunz, line 18: I'd prefer to structure these like the existing specs we have for spans where as much as possible we assert against written out literal json, e.g.:

```
expect(logs).toEqual([
  {
    body: "my error log",
    severity: "info",
    time_unix_nano: logs[0].time_unix_nano,
    ...
  },
  etc.
]);
```

can add sorting during parsing to make sure we always have the same structure and for attributes that we don't care about the values or are dynamic can do something like `time_unix_nano: logs[0].time_unix_nano` to reflect the value back or add helpers like we have with `commonEmbraceSpanAttributes`

integration-tests/helpers/embrace_mock_api.ts
TheMutsi: Yes the other day you told me that instead of env you are going to do something with params or something like that

integration-tests/helpers/log.ts
TheMutsi: But I need remove the item from the response data in the API. because when it compares the json it throw fails

integration-tests/typings/embrace.d.ts
TheMutsi: ok

yarn.lock
TheMutsi: No I though we were only using NPM I wil revert it

integration-tests/helpers/log.ts
jpmunz: if the type doesn't include it when you cast any extra keys should be ignored even if they are included in the raw json

integration-tests/helpers/embrace_mock_api.ts
jpmunz: in integration-tests/wdio.conf.ts there is `parseCLIOptions` that you can add more options to

jpmunz:
closing for now as we revisit the ticket

PR #239, opened 2024-11-01
[EMBR-5417] @embrace-io/react-native-otlp | Create structure for Android native component

facostaembrace:
## Goal

- creating structure for RNEmbraceOTLP Android (solution itself will be added in a separated pr, but having this structure will let Android also run in `main` without disabling otlp component)


packages/react-native-otlp/android/dependencies.gradle
```diff
@@ -0,0 +1,6 @@
+dependencies {
+  api "io.embrace:embrace-android-sdk:6.13.0"
```
facostaembrace, line 2: TBD: to be pulled from package.json

facostaembrace: update: covered

packages/react-native-otlp/android/src/main/java/io/embrace/rnembraceotlp/RNEmbraceOTLPModule.kt
jpmunz, line 1: something to note for the consistency task we have `io/embrace/embracewrapper/EmbraceManagerModule`vs `io/embrace/reactnativetracerprovider/ReactNativeTracerProviderModule` for the other packages currently

packages/react-native-otlp/android/dependencies.gradle
```diff
@@ -0,0 +1,14 @@
...
+def packageJsonFile = file("../package.json")
+
+def jsonSlurper = new JsonSlurper()
+def packageJson = jsonSlurper.parse(packageJsonFile)
+
+dependencies {
+  api "io.embrace:embrace-android-sdk:$packageJson.embrace.androidVersion"
+
+  implementation(platform("io.opentelemetry:opentelemetry-bom:1.43.0"))
+  implementation("io.opentelemetry:opentelemetry-exporter-otlp-http-trace:1.14.0")
```
jpmunz, line 13: on the iOS side GRPC was exposed as well, will that be needed here as a dependency?

jpmunz (approved):
Looks good, for running tests I do think the source files will need to live under react-native-otlp/test-project in order to compile but that can be revisited as part of the next set of PRs

packages/react-native-otlp/android/dependencies.gradle
facostaembrace: I think I've removed the GRPC dependency from the iOS side, I don't see any mention of it (neither in the dependencies nor in the copied/pasted files)

jpmunz: ah ok nm then

PR #245, opened 2024-11-06
[EMBR-5419] Integrate native test flow for custom export config

facostaembrace:
## Goal

- Adding unit tests for otlp component
- Using local package instead of duplicating source files
- TBD: improve/add assertions

PR #249, opened 2024-11-11
[EMBR-5422] Js Unit Tests + Integration Tests

facostaembrace:
## Goal

- renaming test cases (there is no need of having differences between ios/android cases)
- cleaning up what is throwing the Js Layer something there fails (technically the Native layer will always resolve a boolean)
- TBD: will clean up the PR after https://github.com/embrace-io/embrace-react-native-sdk/pull/245 is merged -> Done


packages/react-native-otlp/src/index.ts
```diff
@@ -74,9 +74,8 @@ const initialize = (otlpExporterConfig: OTLPExporterConfig) => {
         sdkConfig,
         otlpExporterConfig,
       );
-    } catch (error) {
-      console.warn(WARN_MESSAGES.error, error);
-      return Promise.reject(false);
+    } catch (_) {
+      return Promise.reject(WARN_MESSAGES.error);
```
facostaembrace, line 78: @jpmunz I've updated this block since I guess it makes sense to reject with at least an error message instead of a boolean value. I think there is no scenario when the code will throw (since it is always resolving the promise with the boolean) but I guess it makes sense to keep the try/catch due to possible exceptions coming from the react native layer

packages/react-native-otlp/src/__tests__/index.test.ts
```diff
@@ -100,6 +105,40 @@ describe("React Native OTLP", () => {
...
+      Platform.OS = "android";
+    });
+
+    it("if it receives the proper configuration (happy path)", async () => {
+      const otlpExporterConfig = {
+        logExporter: {
+          endpoint: "https://example.com/logs/v1",
+          headers: [
+            {key: "Authorization", token: "Bearer token"},
+            {key: "Authorization", token: "Bearer token:v2"},
```
jpmunz, line 119: is it valid to have the same key twice? Possibly headers should be a dictionary actually, but can revisit that separately from this PR

jpmunz (approved):

PR #250, opened 2024-11-12
[EMBR-5618] Unify Native Modules: React Native Tracer Provider + React Native OTLP

facostaembrace:
## Goal

To sync and unify native strategies for OTLP + Tracer Provider packages.
This include test suite + source code, making them compatible with older versions of React Native.
Covering also https://www.notion.so/embraceio/Move-information-from-react-native-tracer-provider-DEVELOPING-md-into-root-file-13c7e3c9985280ecbaa7f982fc1261db


packages/react-native-otlp/package.json
```diff
@@ -4,7 +4,13 @@
   "description": "A React Native wrapper for the Embrace SDK that allows to connect to OTLP endpoints",
   "files": [
     "lib",
-    "android",
+    "android/gradle",
```
jpmunz, line 7: are we sure we need these?
* android/gradle
* android/gradlew
* android/gradlew.bat

the current core package doesn't include them, I think it'd be the customer's app's gradle that would get used when building

packages/react-native-tracer-provider/ios/ReactNativeTracerProvider/ReactNativeTracerProviderModule.swift
```diff
@@ -139,7 +140,7 @@ class ReactNativeTracerProviderModule: NSObject {
   @objc(setupTracer:version:schemaUrl:)
   func setupTracer(name: String, version: String, schemaUrl: String) {
     if tracerProvider == nil {
-      if let isStarted = Embrace.client?.started {
+        if (Embrace.client?.started) != nil {
```
jpmunz, line 143:
```suggestion
      if (Embrace.client?.started) != nil {
```

packages/react-native-tracer-provider/android/build.gradle
```diff
@@ -93,9 +95,8 @@ dependencies {
   // For < 0.71, this will be from the local maven repo
   // For > 0.71, this will be replaced by `com.facebook.react:react-android:$version` by react gradle plugin
   //noinspection GradleDynamicVersion
-  implementation "com.facebook.react:react-native:+"
+  implementation "com.facebook.react:react-android:+"
```
jpmunz, line 98: should this be left as react-native? Based on the comment above it sounds like the plugin is automatically renaming this on newer versions

packages/react-native-tracer-provider/android/build.gradle
jpmunz, line 1: overall for the changes in this file it seems different than what you get by default from `create-react-native-library`, are they required?

scripts/create-native-package.sh
```diff
@@ -0,0 +1,67 @@
+## create package directory under /packages
```
jpmunz, line 1: would it be simpler to start w/ running `create-react-native-library` and work from there?

facostaembrace:
@jpmunz I've polished and cleaned up the solution in favor of boilerplate code coming from `create-react-native-library`, supporting older versions of RN as discussed (mainly supporting local download of RN dependency for Android, keeping `implementation "com.facebook.react:react-native:+"` for source code). Please, have a new look whenever you get the chance

packages/react-native-tracer-provider/android/build.gradle
facostaembrace: Covered. Refactored once again the solution and got it working.

packages/react-native-tracer-provider/android/build.gradle
facostaembrace: Cleaned. Using now mostly what `create-react-native-library` produces when running in --local mode

packages/react-native-otlp/android/build.gradle
```diff
@@ -93,11 +90,11 @@ def kotlin_version = getExtOrDefault("kotlinVersion")
...
-  implementation "com.facebook.react:react-android:+"
+  // For > 0.71, this will be replaced by `com.facebook.react:react-android:$version` by react gradle plugin
+  //noinspection GradleDynamicVersion
+  implementation "com.facebook.react:react-native:+"
   implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
 }
 
-apply from: 'dependencies.gradle'
+// adding Embrace + OpenTelemetry dependencies
+apply from: 'dependencies.gradle'
```
jpmunz, line 100: btw it might still be better to split these for readable but I think I originally only put this in a separate "dependencies.gradle" file because of having the separate native-src/ directory so could consider re-combining

packages/react-native-tracer-provider/android/gradle.properties
```diff
@@ -1,5 +1,9 @@
-TracerProvider_kotlinVersion=1.7.0
-TracerProvider_minSdkVersion=21
-TracerProvider_targetSdkVersion=31
-TracerProvider_compileSdkVersion=31
+android.useAndroidX=true
+
+TracerProvider_packageJsonPath=..
```
jpmunz, line 3: for consistency with `packages/react-native-otlp/android/gradle.properties` might be better to prefix these with `RNEmbraceTracerProvider_`

packages/react-native-otlp/android/build.gradle
```diff
@@ -1,15 +1,15 @@
...
-  def gradle_version = project.properties["RNEmbraceOTLP_gradleVersion"]
 
   repositories {
     google()
     mavenCentral()
   }
 
   dependencies {
-    classpath "com.android.tools.build:gradle:$gradle_version"
+    classpath "com.android.tools.build:gradle:8.6.0"
```
jpmunz, line 11: this is the one bump I'm still not fully sure on, looks like the create library tool outputs it as "7.2.1". Do you know if it has any impact on how the customer's app gets built? That might be hard to dig into to, alternatively could try keeping it as "7.2.1" and building on an app that has a higher version that's above our minimum required

packages/react-native-otlp/android/build.gradle
facostaembrace: yep, I kept this in a separate file for reusability purposes

packages/react-native-otlp/android/build.gradle
facostaembrace: will take a look if I still need to run the `gradle wrapper` commando that would update what it is produced there

facostaembrace: done. I was able to set everything with a minimum of 7.2.1

packages/react-native-otlp/RNEmbraceOTLP.podspec
```diff
@@ -1,22 +1,20 @@
...
 
 Pod::Spec.new do |s|
-  s.name = "RNEmbraceOTLP"
-  s.version = package["version"]
-  s.summary = package["description"]
-  s.homepage = package["homepage"]
-  s.license = package["license"]
-  s.authors = package["author"]
-
-  s.ios.deployment_target = '13.0'
```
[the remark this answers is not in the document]
facostaembrace: it was, because the `create-react-native-library` did not have this min version added. but I think since it was already tested in the iOS build I would re-introduce it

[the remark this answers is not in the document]
jpmunz: ya I think this would be more specific to our minimum requirements dictated by the iOS SDK. Actually I think we'd be fine leaving these off and only including them in the core package's podspec? Maybe a question for the iOS team to confirm but since they are specific to the SDK requirements that is the only package that references it directly so maybe not needed here after all. If we do need them in multiple packages we'll probably want to read from package.json to keep consistent

scripts/create-native-package.sh
jpmunz: can this be removed now? I think all the info got moved to NATIVE_MODULE_DEVELOPING.md ?

packages/react-native-otlp/RNEmbraceOTLP.podspec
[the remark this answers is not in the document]
facostaembrace: yes, the core package already contains the following

```
  s.ios.deployment_target = '13.0'
  s.tvos.deployment_target = '13.0'
  s.swift_version = '5.0'
```

ok, so I'll remove the `s.ios.deployment_target = '13.0'` from the OTLP package since this is behaving as a direct dependency of Core

jpmunz (approved):

PR #251, opened 2024-11-12
[EMBR-5063] Clean/unify native modules: Embrace Core

facostaembrace:
## Goal

Updating Core package following same strategy of otlp/tracer provider package.
Added test-project/android


packages/core/test-project/android/config/detekt/detekt.yml
jpmunz, line 1: is this the same for all 3 packages? Is it possible to move up to the root?

packages/core/test-project/android/app/src/test/java/io/embrace/embracewrappertest/RNEmbraceCoreTest.kt
jpmunz, line 1: nice

packages/core/package.json
```diff
@@ -15,11 +15,13 @@
   },
   "files": [
     "lib",
+    "android/gradle",
     "android/src",
     "android/build.gradle",
+    "android/dependencies.gradle",
     "android/gradle.properties",
-    "ios/RNEmbrace",
```
jpmunz, line 21: with this change `packages/core/ios/RNEmbrace.xcodeproj` should be removed if it's no longer needed to avoid it getting included in the package

packages/core/android/gradle.properties
```diff
@@ -1 +1,11 @@
-emb_android_sdk=6.13.0
\ No newline at end of file
+android.useAndroidX=true
+
+emb_android_sdk=6.13.0
```
jpmunz, line 3: this is no longer needed right since we're reading it from the package json now?

packages/core/android/build.gradle
```diff
@@ -1,45 +1,101 @@
...
+  def kotlin_version = rootProject.ext.has("kotlinVersion") ? rootProject.ext.get("kotlinVersion") : project.properties["RNEmbraceCore_kotlinVersion"]
+
+  repositories {
+    google()
+    mavenCentral()
+  }
+
+  dependencies {
+    classpath "com.android.tools.build:gradle:8.6.0"
+    classpath "io.embrace:embrace-swazzler:$emb_android_sdk"
```
jpmunz, line 12: ah I see `emb_android_sdk` is still being read here, if possible we should have it grabbed from package.json as well, if not may need to restore part of `packages/core/update_android_version.ts` since it was previously keeping the value for it in sync

packages/core/android/build.gradle
```diff
@@ -1,45 +1,101 @@
...
-    defaultConfig {
-        minSdkVersion 16
-        targetSdkVersion 26
-        versionCode 1
-        versionName "1.0"
-        ndk {
-            abiFilters "armeabi-v7a", "x86"
-        }
+  if (supportsNamespace()) {
+    namespace "io.embrace.embracewrapper"
```
jpmunz, line 53: we still seem to flip in a couple spots between calling it embrace core vs embrace wrapper, should try and unify but more minor though

packages/core/android/build.gradle
facostaembrace: I'm in favor of getting rid of `emb_android_sdk` but this change may impact in other places as well like the install script + related tests. I guess I would vote for this to get `packages/core/update_android_version.ts` back and clean it up in a next iteration

packages/core/test-project/android/config/detekt/detekt.yml
facostaembrace: yes, I filled a ticket for it https://www.notion.so/embraceio/Move-detekt-config-for-Android-to-the-root-1447e3c998528030a7d6ff4763134bee

integration-tests/basic-test-app/ios/Podfile.lock
```diff
@@ -1500,9 +1500,28 @@ PODS:
     - React-logger (= 0.74.1)
     - React-perflogger (= 0.74.1)
     - React-utils (= 0.74.1)
-  - RNEmbrace (5.0.2):
+  - RNEmbraceCore (5.0.2):
```
jpmunz, line 1503: hmm I'm curious on this change, it looks like previously the dependencies of this pod was just:
```
  - RNEmbrace (5.0.2):
    - EmbraceIO (= 6.5.1)
    - React-Core
```

might be expected but do you know what changes to the package caused all the other ones to be pulled in? (FWIW seems like TracerProvider, etc. also have the same larger list of dependencies)

jpmunz: oh I guess it's due to this in the podspec:
```

  # Use install_modules_dependencies helper to install the dependencies if React Native version >=0.71.0.
  # See https://github.com/facebook/react-native/blob/febf6b7f33fdb4904669f99d795eba4c0f95d7bf/scripts/cocoapods/new_architecture.rb#L79.
  if respond_to?(:install_modules_dependencies, true)
    install_modules_dependencies(s)
  else
    s.dependency "React-Core"
```

which seems expected and the way RN wants to do things on newer versions?

packages/core/ios/RNEmbraceCore/SpanRepository.swift
```diff
@@ -6,9 +6,9 @@ import OpenTelemetryApi
 private let MAX_STORED_SPANS = 10000
 
 class SpanRepository {
-    private let activeSpansQueue = DispatchQueue(label: "io.embrace.RNEmbrace.activeSpans",
```
jpmunz, line 9: btw I think all this will be transparent to the user but let's remember to do a test during QA for next release where we try upgrading an app from a previous version to one with the new name and see if anything weird happens

packages/core/ios/RNEmbraceCore.xcodeproj/project.pbxproj
jpmunz, line 1: is this file actually needed?

integration-tests/basic-test-app/ios/Podfile.lock
facostaembrace: yes, it's expected. this is because I updated the previous .podspec file to use the same structure + dependencies as what `create-react-native-library` creates. I don't see a reason why we don't update this podspec to be in sync with the other packages

packages/core/package.json
facostaembrace: done

packages/core/android/build.gradle
facostaembrace: done

packages/core/ios/RNEmbraceCore.xcodeproj/project.pbxproj
facostaembrace: done

jpmunz (approved):

PR #256, opened 2024-11-18
Add default bundle path for ios

TheMutsi:
## Goal

Grab the default main js bundle for react native iOS

### DIFF
--- a/packages/core/ios/RNEmbrace/BundleID.swift
+++ b/packages/core/ios/RNEmbrace/BundleID.swift
@@ -55,18 +55,23 @@ func computeBundleID(path: String) throws -> BundleID {
             return BundleID(id: last.id, cached: true)
         }
     }
+    
+    do {
+        let fileData = try Data(contentsOf: URL(fileURLWithPath: path))
 
-    let fileContents = try String(contentsOf: URL(fileURLWithPath: path))
+        // https://stackoverflow.com/a/56578995
+        let bundleID = Insecure.MD5.hash(data: fileData).map {
+            String(format: "%02hhx", $0)
+        }.joined()
+        
+        last.computedAt = Date()
+        last.path = path
+        last.id = bundleID
+        last.store()
 
-    // https://stackoverflow.com/a/56578995
-    let bundleID = Insecure.MD5.hash(data: Data(fileContents.utf8)).map {
-        String(format: "%02hhx", $0)
-    }.joined()
-
-    last.computedAt = Date()
-    last.path = path
-    last.id = bundleID
-    last.store()
-
+    } catch {
+        throw ComputeBundleIDErrors.parseError
+    }
+    
     return BundleID(id: last.id, cached: false)
 }
--- a/packages/core/ios/RNEmbrace/EmbraceManager.m
+++ b/packages/core/ios/RNEmbrace/EmbraceManager.m
@@ -6,6 +6,9 @@ RCT_EXTERN_METHOD(setJavaScriptBundlePath:(NSString *)path
                   resolver:(RCTPromiseResolveBlock)resolve
                   rejecter:(RCTPromiseRejectBlock)reject)
 
+RCT_EXTERN_METHOD(getDefaultJavaScriptBundlePath:(RCTPromiseResolveBlock)resolve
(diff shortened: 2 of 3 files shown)


packages/core/src/index.ts
```diff
@@ -125,6 +125,13 @@ const initialize = async ({
   // Only attempt to check for CodePush bundle URL in release mode. Otherwise CodePush will throw an exception.
   // https://docs.microsoft.com/en-us/appcenter/distribution/codepush/react-native#plugin-configuration-ios
   if (!__DEV__) {
+    const bundleJs =
+      await NativeModules.EmbraceManager.getDefaultJavaScriptBundlePath();
+
+    if (bundleJs) {
+      await NativeModules.EmbraceManager.setJavaScriptBundlePath(bundleJs);
```
TheMutsi, line 132: I await this since I want Codepush to be the last to be applied

packages/core/src/index.ts
```diff
@@ -125,6 +125,13 @@ const initialize = async ({
   // Only attempt to check for CodePush bundle URL in release mode. Otherwise CodePush will throw an exception.
   // https://docs.microsoft.com/en-us/appcenter/distribution/codepush/react-native#plugin-configuration-ios
   if (!__DEV__) {
+    const bundleJs =
+      await NativeModules.EmbraceManager.getDefaultJavaScriptBundlePath();
```
jpmunz, line 129: this will break on Android since the method isn't implemented there

packages/core/src/index.ts
jpmunz: In the case of Codepush being implemented does this mean we'd be computing the bundle ID twice?

An alternative might be to stick this default behaviour in the else case here when codepush isn't setup: https://github.com/embrace-io/embrace-react-native-sdk/blob/9e999556384b761774cc678d232c07992d27e0b0/packages/core/ios/RNEmbrace/EmbraceManager.swift#L224, that could also address this being an ios-only special case

packages/core/src/index.ts
```diff
@@ -125,7 +125,17 @@ const initialize = async ({
   // Only attempt to check for CodePush bundle URL in release mode. Otherwise CodePush will throw an exception.
   // https://docs.microsoft.com/en-us/appcenter/distribution/codepush/react-native#plugin-configuration-ios
   if (!__DEV__) {
-    NativeModules.EmbraceManager.checkAndSetCodePushBundleURL();
```
jpmunz, line 128: with these being switched to using `await` they now have the possibility of raising an exception if the promises are rejected, probably should wrap this whole section in a try/catch to avoid crashing if any of them fail

packages/core/ios/RNEmbrace/BundleID.swift
```diff
@@ -55,18 +55,23 @@ func computeBundleID(path: String) throws -> BundleID {
...
+            String(format: "%02hhx", $0)
+        }.joined()
+        
+        last.computedAt = Date()
+        last.path = path
+        last.id = bundleID
+        last.store()
 
+    } catch {
+        throw ComputeBundleIDErrors.parseError
```
jpmunz, line 73: this symbol doesn't appear to be defined, above I only see the 'emptyPath' error:
```
enum ComputeBundleIDErrors: Error {
    case emptyPath
}
```

packages/core/src/index.ts
```diff
@@ -125,7 +125,23 @@ const initialize = async ({
...
+    try {
+      const isCodePushPresent =
+        await NativeModules.EmbraceManager.checkAndSetCodePushBundleURL();
+
+      if (!isCodePushPresent && Platform.OS === "ios") {
```
jpmunz, line 132:
```suggestion
      // On Android the Swazzler stores the computed bundle ID as part of the build process and the SDK is able to
      // read it at run time. On iOS however we don't retain this value so we either need to get it from the Code Push
      // bundle or, if that isn't enabled, try and get it from the default bundle path
      if (!isCodePushPresent && Platform.OS === "ios") {
```

jpmunz (approved):
looks good, I just added a comment for the special case so that we have the context going forward

jpmunz (approved):

PR #261, opened 2024-11-20
[EMBR-5298] Dave React Native iOS: TypeError crashes

TheMutsi:
## Goal

Handle render error adding more information in the timeline related to the component  stack

## Testing

Generating a crash that happens in the JSX 

NEW LINKS
With Component Error
https://dash.embrace.io/app/BAyj6/grouped_sessions/last_48_hour/F92967E0C1304D57A5AE9DF4654EFB26/583010CA1F394AAD8DD2CFE602C46FBB@/583010CA1F394AAD8DD2CFE602C46FBB?first=0C25EDA245994C38AFC0C2DB0A407A42&last=6E613C5C30F540DF871C10792CAD4977&source=session

https://dash.embrace.io/app/BAyj6/grouped_sessions/last_48_hour/F92967E0C1304D57A5AE9DF4654EFB26/583010CA1F394AAD8DD2CFE602C46FBB@/583010CA1F394AAD8DD2CFE602C46FBB?first=0C25EDA245994C38AFC0C2DB0A407A42&last=6E613C5C30F540DF871C10792CAD4977&source=session

Without Component Error
https://dash.embrace.io/app/Jso7A/grouped_sessions/last_1_hour/3E10FDEFAF7A45DFB41E008A1FE2A7B0/EBAD1BC88BA64DBAA34FBC320B5F6FED@/EBAD1BC88BA64DBAA34FBC320B5F6FED?first=EAF3E955B18546829AC1C1E05C710047&last=EBAD1BC88BA64DBAA34FBC320B5F6FED&source=session
https://dash.embrace.io/app/BAyj6/grouped_sessions/last_1_hour/82778FC879CC4A8F9E2F43F01BA9CA69/7D6D71AB99614F9EBCF445FCF5AE6CA4@/7D6D71AB99614F9EBCF445FCF5AE6CA4?first=80230CE9A55D451DA9D5E6C08F1C0158&last=9115B1ED868B40F1A33A05F9232981E6&source=session


facostaembrace (commented):
Can we add some js tests covering this new flow?

packages/core/src/index.ts
```diff
@@ -17,6 +17,18 @@ interface Properties {
   [key: string]: string;
 }
 
+class UIError extends Error {
```
facostaembrace, line 20: actually, if we are not instantiating this class (so, we are not doing any `new UIError(...)`), do we need it?

if it is only about having the type, could be possible to set something like

```
interface UIError extends Error {
  componentStack: string;
}
```

packages/core/src/index.ts
```diff
@@ -277,10 +298,15 @@ export const logHandledError = (
 ): Promise<boolean> => {
   if (error instanceof Error) {
     const {stack, message} = error;
+    let finalStack = stack;
```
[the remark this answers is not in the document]
jpmunz: actually if `logComponenError` was written to return early if there wasn't a component stack available then maybe we don't even need the if check here, can just be a `logComponentError(error)` one liner and we keep all the logic in the helper

packages/core/src/index.ts
```diff
@@ -277,10 +298,15 @@ export const logHandledError = (
...
+    if ("componentStack" in error) {
+      const uiError = error as UIError;
```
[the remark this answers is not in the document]
TheMutsi: I tried that, TS doesn't like that way... so I had to put as it is now

packages/core/src/utils/ErrorUtil.ts
```diff
@@ -22,4 +32,24 @@ const generateStackTrace = (): string => {
   return err.stack || "";
 };
 
-export {handleGlobalError, generateStackTrace};
+const isJSXError = (error: Error | UIError): error is UIError => {
```
jpmunz, line 35: I would put all these additions in a new file to make back porting easier

packages/core/src/index.ts
```diff
@@ -29,14 +36,16 @@ const ERROR = "error";
 const noOp = () => {};
 
 // will cover unhandled errors, js crashes
-const handleError = async (error: Error, callback: () => void) => {
+const handleError = async (error: Error | UIError, callback: () => void) => {
```
jpmunz, line 39: shouldn't be needed I don't think, the new type can be handled inside `logIfComponentError`

packages/core/src/index.ts
```diff
@@ -272,7 +281,7 @@ export const logError = (message: string): Promise<boolean> => {
 };
 
 export const logHandledError = (
-  error: Error,
+  error: Error | UIError,
```
jpmunz, line 284: similar

packages/core/src/utils/ErrorUtil.ts
```diff
@@ -22,4 +32,24 @@ const generateStackTrace = (): string => {
...
+  return "componentStack" in error;
+};
+
+export const logIfComponentError = (
+  error: Error | UIError,
+  properties?: Properties,
+): Promise<boolean> => {
+  if (!isJSXError(error) || error.componentStack !== "") {
+    return createFalsePromise();
```
jpmunz, line 44: I think we can move away from these helpers, I believe just `Promise.resolve(false)` is equivalent

packages/core/src/utils/ErrorUtil.ts
```diff
@@ -22,4 +32,24 @@ const generateStackTrace = (): string => {
...
-export {handleGlobalError, generateStackTrace};
+const isJSXError = (error: Error | UIError): error is UIError => {
+  return "componentStack" in error;
+};
+
+export const logIfComponentError = (
+  error: Error | UIError,
+  properties?: Properties,
+): Promise<boolean> => {
+  if (!isJSXError(error) || error.componentStack !== "") {
```
jpmunz, line 43: shouldn't this be `error.componentStack === ""` ?

packages/core/src/index.ts
```diff
@@ -401,21 +405,5 @@ export const applyNetworkInterceptors = (
   return createTruePromise();
 };
 
-const createFalsePromise = (): Promise<boolean> => {
```
jpmunz, line 404: I'd leave these in place for the purposes of this PR, can revisit refactoring or replacing with `Promise.resolve` where they get used separately

packages/core/src/utils/ComponentError.ts
jpmunz, line 1: let's add a larger comment to this file explaining the usecase and including examples of what the component tree looks like and why just having the normal JS stack trace isn't as helpful

packages/core/src/__tests__/error.test.ts
```diff
@@ -0,0 +1,56 @@
...
+jest.mock("react-native", () => ({
+  NativeModules: {
+    EmbraceManager: {
+      logHandledError: (
+        message: string,
+        componentStack: string,
+        params: object,
+      ) => {
+        mockLogHandledError(message, componentStack, params);
+        return true;
```
jpmunz, line 14: technically the real implementation would be returning something like `Promise.resolve(true)`

packages/core/src/index.ts
TheMutsi: sure

packages/core/src/utils/ComponentError.ts
```diff
@@ -0,0 +1,54 @@
...
+ * - This function is complementary to other logging mechanisms that capture more
+ *   detailed stack traces, including line and column numbers.
+ * - By default, in release builds, the screen names in the `componentStack` appear as
+ *   `Unknown`. To retain meaningful screen names in release builds, you must configure
+ *   Metro to preserve the names of React components during minification. For detailed
+ *   instructions, refer to https://docs.embrace.io/docs/react-native/features/tracking-error-component/.
+ *
+ * Example `componentStack` trace:
+ * ```
+ * at undefined (in App)
```
jpmunz, line 31: same question from the docs PR, will these always be "at undefined"? If so we may want to remove them so that we log a cleaner output

TheMutsi: We chan change it to: "at anonymous"

jpmunz: I'd like to understand a bit more where that is coming from, is it always going to be "at undefined" or are there cases where it will show different information?

packages/core/src/utils/ComponentError.ts
```diff
@@ -0,0 +1,58 @@
...
+ * if the error was not a React Native rendering error or the `componentStack` was empty.
+ */
+const logIfComponentError = (error: Error): Promise<boolean> => {
+  if (!isJSXError(error) || error.componentStack === "") {
+    return Promise.resolve(false);
+  }
+  const {message, componentStack} = error;
+
+  const componentStackShrinked = componentStack
+    .replace(/at UNdefined|\(|\)/g, "")
```
jpmunz, line 48: should define the regex as a constant outside so that it doesn't need to be reevaluated on each call

for the regex itself what is the function of the `|` characters? You could try this version: `.replace(/at undefined \((.*)\)/g, "$1")`

packages/core/src/__tests__/initialize.test.ts
```diff
@@ -134,6 +144,27 @@ describe("Android: initialize", () => {
...
+    ErrorUtils.getGlobalHandler = previousHandler;
+    const result = await initialize({patch: testValue});
+    expect(result).toBe(true);
+
+    const generatedGlobalErrorFunc = handleGlobalError(
+      previousHandler,
+      logIfComponentError,
+    );
+    const componentError = new Error("Test") as ComponentError;
+    componentError.componentStack = "in SomeScreen/n in SomeOtherScreen";
```
jpmunz, line 159: should include the "at undefined" stuff in the input and then verify it is removed in the output

packages/core/src/__tests__/error.test.ts
```diff
@@ -0,0 +1,56 @@
...
+
+jest.mock("react-native", () => ({
+  NativeModules: {
+    EmbraceManager: {
+      logHandledError: (
+        message: string,
+        componentStack: string,
+        params: object,
+      ) => {
+        mockLogHandledError(message, componentStack, params);
```
facostaembrace, line 13: the `mockLogHandledError` itself could return the `true`. I believe something like

```
const mockLogHandledError = jest.fn().mockReturnValue(true);
```

and after something like

```
...
logHandledError: (
        message: string,
        componentStack: string,
        params: object,
      ) => mockLogHandledError(message, componentStack, params),
...

```

and similar in other places

packages/core/src/utils/ComponentError.ts
TheMutsi: This
`/at undefined \((.*)\)/g` delete the whole text, I receive ""

jpmunz (approved):

jpmunz (approved):

PR #269, opened 2024-11-28
[EMBR-5717] Android - Support for w3c traceparent header

facostaembrace:
## Goal

- Introducing support for w3ctraceparent header
- Unit tests

### DIFF
--- a/packages/core/android/src/main/java/io/embrace/rnembracecore/EmbraceManagerModule.java
+++ b/packages/core/android/src/main/java/io/embrace/rnembracecore/EmbraceManagerModule.java
@@ -9,14 +9,11 @@ import com.facebook.react.bridge.ReadableMap;
 import com.facebook.react.bridge.ReadableArray;
 
 import com.facebook.react.bridge.Promise;
-import com.facebook.react.bridge.WritableNativeMap;
-import com.facebook.react.bridge.WritableMap;
-
-import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.Map;
 import java.util.HashMap;
 
+import io.embrace.android.embracesdk.internal.network.http.InternalNetworkApiImpl;
 import io.embrace.android.embracesdk.spans.ErrorCode;
 import java.util.ArrayList;
 import java.util.List;
@@ -27,6 +24,7 @@ import io.embrace.android.embracesdk.Embrace;
 import io.embrace.android.embracesdk.Severity;
 import io.embrace.android.embracesdk.network.EmbraceNetworkRequest;
 import io.embrace.android.embracesdk.network.http.HttpMethod;
+import io.embrace.android.embracesdk.internal.EmbraceInternalInterface;
 
 public class EmbraceManagerModule extends ReactContextBaseJavaModule {
     private ReactApplicationContext context;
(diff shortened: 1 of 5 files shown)


packages/core/test-project/android/app/src/test/java/io/embrace/rnembracecoretest/RNEmbraceCoreTest.kt
```diff
@@ -4,36 +4,157 @@ import android.os.SystemClock
...
+                    endTime,
+                    errorType,
+                    errorMessage,
+                    promise,
+                )
+
+                verify(mockEmbrace).generateW3cTraceparent()
+
+                mockedStatic.verify {
+                    EmbraceNetworkRequest.fromIncompleteRequest(
```
jpmunz, line 143: in the interest of mocking less I think we should setup a log exporter on an actual embrace SDK instance similar to how the otlp package does it and then verify that the traceparent is set based on the attributes of the log that was generated by `logNetworkClientError`

packages/core/android/src/main/java/io/embrace/rnembracecore/EmbraceManagerModule.java
```diff
@@ -395,21 +395,26 @@ public void logNetworkRequest(String url,
...
                     url,
                     HttpMethod.fromInt(method),
                     st,
                     et,
                     bytesSent.intValue(),
                     bytesReceived.intValue(),
-                    statusCode.intValue()
+                    statusCode.intValue(),
+                    null,
+                    w3cTraceparent,
```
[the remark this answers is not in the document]
facostaembrace: covered

packages/core/android/src/main/java/io/embrace/rnembracecore/EmbraceManagerModule.java
```diff
@@ -571,6 +578,15 @@ public void recordCompletedSpan(String name, Double startTimeMs, Double endTimeM
         }
     }
 
+    public boolean isNetworkSpanForwardingEnabled() {
+        InternalNetworkApiImpl networkApi = new InternalNetworkApiImpl();
```
jpmunz, line 582: would be good to avoid creating this instance on each call, either could try: `EmbraceInternalApi.getInstance().internalInterface.isNetworkSpanForwardingEnabled()` or if that doesn't work saving the instance on the class

facostaembrace: I was able to move the instance outside of the method only. I'm not sure if the SDK exposes the EmbraceInternalApi, or at least I couldn't use it. We can discuss about it during standup

packages/react-native-tracer-provider/test-project/android/app/build.gradle.kts
```diff
@@ -68,6 +68,12 @@ android {
     buildToolsVersion = rootProject.ext["buildToolsVersion"] as String
     compileSdk = rootProject.ext["compileSdkVersion"] as Int
 
+    testOptions {
+        unitTests {
+            isReturnDefaultValues = true
```
jpmunz, line 73: is this required for packages/react-native-otlp as well?

PR #270, opened 2024-11-29
Refactor error handler

TheMutsi:
## Goal

Refactor error handler to remove logic from the /core/index.ts


packages/core/src/__tests__/error.test.ts
```diff
@@ -1,8 +1,22 @@
+import {
+  generateStackTrace,
+  handleError,
+  handleGlobalError,
+  STACK_LIMIT,
+} from "../utils/ErrorUtil";
 import {ComponentError, logIfComponentError} from "../utils/ComponentError";
 
+const mockErrorCallback = jest.fn();
+const mockLogUnhandledJSException = jest.fn();
```
facostaembrace, line 10: this could be improved to 
```
const mockLogUnhandledJSException = jest.fn().mockReturnValue(Promise.resolve(true))
```

packages/core/src/__tests__/error.test.ts
```diff
@@ -11,7 +25,16 @@ jest.mock("react-native", () => ({
...
         mockLogHandledError(message, componentStack, params);
-        return true;
+        return Promise.resolve(true);
+      },
+      logUnhandledJSException: (
+        name: string,
+        message: string,
+        errorType: string,
+        stack: string,
+      ) => {
```
facostaembrace, line 35: and here it can be an inline return for simplicity

```
logUnhandledJSException: (
        name: string,
        message: string,
        errorType: string,
        stack: string,
      ) => mockLogUnhandledJSException(name, message, errorType, stack)
```

packages/core/src/__tests__/error.test.ts
```diff
@@ -41,16 +64,138 @@ describe("Component Error", () => {
...
+    expect(mockLogUnhandledJSException).toHaveBeenCalledWith(
+      name,
+      message,
+      errorType,
+      stTruncated,
+    );
+  });
+  test("Error handler should not call callback", async () => {
+    const customError = {
+      name: "Not Eror",
```
facostaembrace, line 119:
```suggestion
      name: "Not an Error",
```

packages/core/src/__tests__/error.test.ts
```diff
@@ -41,16 +64,138 @@ describe("Component Error", () => {
...
+      true,
+    );
+    jest.runAllTimers();
+    expect(mockPreviousHandler).toHaveBeenCalledWith(customError, true);
+  });
+  test("Generate Stack", async () => {
+    jest.useFakeTimers();
+
+    const customError = new Error("Ups!");
+    const customHandleError = (_: Error, callback: () => void) => {
```
facostaembrace, line 189:
```suggestion
    const handleCustomError = (_: Error, callback: () => void) => {
```

jpmunz:
closing for now as we revisit the ticket

PR #271, opened 2024-11-29
[EMBR-5717] iOS - Support for w3c traceparent header

facostaembrace:
## Goal

Introducing support for w3ctraceparent
Unit tests


packages/core/ios/EmbraceManager.swift
```diff
@@ -21,83 +22,94 @@ private let EVENT_ATTRIBUTES_KEY = "attributes"
...
         } else {
             reject("error", "Unable to retrieve JS bundle path", nil)
         }
     }
+
     @objc
     func isStarted(_ resolve: @escaping RCTPromiseResolveBlock, rejecter reject: @escaping RCTPromiseRejectBlock) {
         if let embraceStarted = Embrace.client?.started {
-        resolve(embraceStarted)
+            resolve(embraceStarted)
```
jpmunz, line 73: you can probably close https://www.notion.so/embraceio/Re-run-swiftlint-over-the-repo-1287e3c99852802390a4ed824e537726?pvs=4 once this is merged as well as this reformatting was some existing debt

packages/core/ios/EmbraceManager.swift
```diff
@@ -21,83 +22,94 @@ private let EVENT_ATTRIBUTES_KEY = "attributes"
...
 
                     let servicesBuilder = CaptureServiceBuilder().addDefaults()
-                    if config.disableAutomaticViewCapture {
+                    if self.config.disableAutomaticViewCapture {
                             servicesBuilder.remove(ofType: ViewCaptureService.self)
                     }
 
+                    if self.config.disableNetworkSpanForwarding {
+                        servicesBuilder.add(.urlSession(options:
+                                                            URLSessionCaptureService.Options(injectTracingHeader: false, requestsDataSource: nil)))
```
jpmunz, line 100: just want to confirm my understanding here, so is it the case that by default with `CaptureServiceBuilder().addDefaults()` we get a  `URLSessionCaptureService` that has injecting tracing headers on? And then this replaces that service rather than adding a 2nd one?

packages/core/ios/EmbraceManager.swift
```diff
@@ -439,14 +451,19 @@ class EmbraceManager: NSObject {
...
         }
 
-        Embrace.client?.recordCompletedSpan(name: createNetworkSpanName(url: url, httpMethod: httpMethod),
-                                            type: SpanType.networkRequest, parent: nil,
-                                            startTime: dateFrom(ms: startInMillis), endTime: dateFrom(ms: endInMillis),
-                                            attributes: attributeStringsFrom(dict: attributes as NSDictionary),
-                                            events: eventsFrom(array: []),
-                                            errorCode: nil
-        )
+        var embraceOtel = EmbraceOTel()
```
jpmunz, line 454: should probably do `Embrace.client?.buildSpan` instead since that is how we're invoking it in the other methods

packages/core/ios/EmbraceManager.swift
```diff
@@ -439,14 +451,19 @@ class EmbraceManager: NSObject {
...
+        var spanBuilder = embraceOtel
```
jpmunz, line 455: since `spanBuilder` isn't otherwise being used you could do grab the `span` directly:

```
let span =  Embrace.client?.buildSpan(...).setStartTime(...).startSpan()
```

packages/core/ios/EmbraceManager.swift
```diff
@@ -808,4 +832,9 @@ class EmbraceManager: NSObject {
 
         resolve(true)
     }
+
+    func injectW3cTraceparent(span: OpenTelemetryApi.Span) {
```
jpmunz, line 836: I think this would work without fully qualifying it:
```suggestion
    func injectW3cTraceparent(span: Span) {
```

packages/core/test-project/ios/RNEmbraceCoreTests/RNEmbraceCoreTests.swift
```diff
@@ -731,11 +735,13 @@ class EmbraceManagerTests: XCTestCase {
         XCTAssertEqual(exportedSpans[0].attributes["http.response.body.size"]!.description, "2000")
         XCTAssertEqual(exportedSpans[0].attributes["http.request.body.size"]!.description, "1000")
         XCTAssertEqual(exportedSpans[0].attributes["http.response.status_code"]!.description, "200")
+        XCTAssertNotNil(exportedSpans[0].attributes["emb.w3c_traceparent"])
```
jpmunz, line 738: is there a way to test the case where the config is off and we can verify the attribute isn't set? Might not be possible if that involves starting the SDK from scratch again since we want to avoid that in testing

facostaembrace: confirmed by Ari, there is no way of updating the config through the code (it's inmutable). not sure if there is other way of mocking the `config` variable and make it return a falsy value for the `disableNetworkSpanForwarding` key

packages/core/ios/EmbraceManager.swift
facostaembrace: because of this comment I've dug into this scenario. As per what I read in the ios sdk repo is [here](https://github.com/embrace-io/embrace-apple-sdk/blob/7ae30ac688c6d3cf8c19e923833812ce10457326/Sources/EmbraceIO/Capture/CaptureServiceBuilder.swift#L40) I think yes, it would add a 2nd service so I've refactored it a little bit to be sure the service is not added twice. I'll perform some manual tests and see if I can introduce tweaks that tests this situation during unit tests if possible

packages/core/test-project/ios/RNEmbraceCoreTests/RNEmbraceCoreTests.swift
jpmunz: I wouldn't spend too much time on it, it's something we can punt more to an integration test and spin up an app with a different config

for the purposes of this testing I guess could maybe move the `if` check inside the method and pass in the config as a parameter, could then have a unit test just on that method and confirm the attribute being added or not...though that may not add too much value

packages/core/ios/EmbraceManager.swift
```diff
@@ -21,83 +22,98 @@ private let EVENT_ATTRIBUTES_KEY = "attributes"
...
+
+                        // manually adding the URLSessionCaptureService
+                        servicesBuilder.add(.urlSession(options: urlSessionServiceOptions))
                     }
 
-                    let servicesBuilder = CaptureServiceBuilder().addDefaults()
-                    if config.disableAutomaticViewCapture {
-                            servicesBuilder.remove(ofType: ViewCaptureService.self)
+                    // adding defaults
+                    servicesBuilder.addDefaults()
```
jpmunz, line 100: hmm isn't this still adding the urlSessionService again though if `self.config.disableNetworkSpanForwarding` was set? Or is `addDefaults` checking that it doesn't re-add a service that is already on the builder?

facostaembrace: as per what I understood from the message I shared before the service won't be replaced unless I'm wrong. @ArielDemarco @NachoEmbrace could you please take a look? is this the right way of implementing this cenario? is there any improvement I can work on?

ArielDemarco: If you already set a specific `CaptureService` (like `URLSessionCaptureService`) the `addDefaults` method won't override it. This method checks if the service was already added.

packages/core/ios/EmbraceManager.swift
```diff
@@ -21,83 +22,98 @@ private let EVENT_ATTRIBUTES_KEY = "attributes"
...
+
+                    let servicesBuilder = CaptureServiceBuilder()
+
+                    if self.config.disableNetworkSpanForwarding {
+                        // allowing to disable NSF by code
+                        let urlSessionServiceOptions = URLSessionCaptureService.Options(injectTracingHeader: false, requestsDataSource: nil)
+
+                        // manually adding the URLSessionCaptureService
+                        servicesBuilder.add(.urlSession(options: urlSessionServiceOptions))
                     }
```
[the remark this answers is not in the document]
facostaembrace: definitely yes, I just wanted to make sure I only do something "custom" in case the setting is disabling the feature. but since you confirmed everything is good I def prefer this way

jpmunz (approved):

packages/core/ios/EmbraceManager.swift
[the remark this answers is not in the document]
facostaembrace: done

PR #277, opened 2024-12-02
EMBR-5820 update core package gradle values

jpmunz:
@facostaembrace it appears that a few of the issues from the ticket were already fixed in the core package's gradle files, I think these were the only ones remaining let me know if you see something I've missed

### DIFF
--- a/packages/core/android/build.gradle
+++ b/packages/core/android/build.gradle
@@ -8,7 +8,7 @@ buildscript {
   }
 
   dependencies {
-    classpath "com.android.tools.build:gradle:7.2.1"
+    classpath "com.android.tools.build:gradle:7.4.2"
     // NOTE: to pull `emb_android_sdk` version from package.json instead in a future iteration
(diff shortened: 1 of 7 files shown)


packages/core/android/build.gradle
```diff
@@ -8,7 +8,7 @@ buildscript {
   }
 
   dependencies {
-    classpath "com.android.tools.build:gradle:7.2.1"
+    classpath "com.android.tools.build:gradle:7.4.2"
```
facostaembrace, line 11: I think this update should also be extended to `react-native-tracer-provider` and `react-native-otlp` packages, right?

and maybe we can add a mention to this in NATIVE_MODULE_DEVELOPING.md, since the `create-react-native-library` by default implements the `com.android.tools.build:gradle:7.2.1` and `https\://services.gradle.org/distributions/gradle-7.4.2-bin.zip`

jpmunz: yep makes sense, updated the packages and the documentation

facostaembrace (approved):

PR #278, opened 2024-12-02
EMBR-5733 create apps for integration testing in a separate directory

jpmunz:
Previously for integration tests app were being created under `embrace-react-native-sdk/integration-tests`, however this caused conflicts with the test apps' dependencies since they were looking into the `node_modules/` folders above them. Change here is to create the apps in an isolated temporary directory

### DIFF
--- a/integration-tests/browserstack.conf.ts
+++ b/integration-tests/browserstack.conf.ts
@@ -111,7 +111,7 @@ exports.config.capabilities.forEach(function (caps) {
  */
 
 // The number of times to retry the entire specfile when it fails as a whole
-exports.config.specFileRetries = 1;
+exports.config.specFileRetries = 2;
 
 // Delay in seconds between the spec file retry attempts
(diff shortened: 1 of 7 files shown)


jpmunz:
Initial run looks better than `main` though there are still failures:
https://github.com/embrace-io/embrace-react-native-sdk/actions/runs/12128946081

Given it was only the older versions of RN that failed I'm inclined to think these are actual failures rather than flakiness with the integration test setup, failures were:
* rn 71 + 72 on android
  * `Detected multiple Kotlin daemon sessions at build/kotlin/sessions`
* rn 71 + 72 on ios
  * `SyntaxError: /Users/runner/work/_temp/rn-apps/rn71/node_modules/expo-router/node_modules/react-native/Libraries/vendor/emitter/EventEmitter.js: Unexpected token, expected "]" (39:5)`

will try locally and see if I can reproduce

facostaembrace (approved):
lgtm!

PR #287, opened 2024-12-09
[EMBR-4929] Make JS stack traces optional for warning and error logs

facostaembrace:
## Goal

- Making optional stacktraces from `error`/`warning` logs (but enabled as default)
- Improving restrictions for `info` logs (should never send stacktraces)
- Cleanup


packages/core/scripts/__tests__/__mocks__/android/buildWithSwazzler.gradle
jpmunz, line 1: can be reverted, we have a ktlo task to address this whitespace

packages/core/src/index.ts
```diff
@@ -256,46 +254,74 @@ export const clearAllUserPersonas = (): Promise<boolean> => {
...
+  const stackTrace =
+    // `LogSeverity.INFO` are not supposed to send stack traces
+    // this is also restricted in the Native layers
+    includeStacktrace && severity !== LogSeverity.INFO
+      ? generateStackTrace()
+      : "";
+
+  if (properties === null) {
+    console.warn(
+      "[Embrace] `properties` is null. It should be an object of type `Properties`. Native layer will ignore this value.",
```
jpmunz, line 277: should we clarify that it could be also be `undefined`?

packages/core/src/__tests__/embrace.test.ts
```diff
@@ -84,7 +80,7 @@ jest.mock("../EmbraceManagerModule", () => ({
     logHandledError: (
       message: string,
       stackTrace: string,
-      properties?: Properties,
+      properties?: LogProperties,
```
jpmunz, line 83: not sure if it matters but technically `properties: LogProperties` would match the real method's signature

jpmunz (approved):

packages/core/src/index.ts
facostaembrace: if `properties` is passed specifically as `undefined` then it will take the default value, that is an empty object `{}` so it won't take any effect

packages/core/src/__tests__/embrace.test.ts
facostaembrace: ah, yeap. let fix it

packages/core/src/index.ts
jpmunz: ah right

PR #299, opened 2024-12-11
fix span testing screen

jpmunz:
Was using this method wrong, the result of startView if it's successful is an id that should then be passed to `endView`

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -301,7 +301,7 @@ export const logHandledError = (
   return createFalsePromise();
 };
 
-export const startView = (view: string): Promise<boolean> => {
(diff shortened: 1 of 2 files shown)


packages/core/src/index.ts
```diff
@@ -301,7 +301,7 @@ export const logHandledError = (
   return createFalsePromise();
 };
 
-export const startView = (view: string): Promise<boolean> => {
+export const startView = (view: string): Promise<string | boolean> => {
```
jpmunz, line 304: would this be a breaking change for anyone consuming this and having used the result as a simple boolean or would typescript resolve the union in that case? If it is breaking we can maybe wait till 6.0 to change this as we'll probably want to remove these methods altogether in favour of working with spans directly

PR #302, opened 2024-12-13
(feature): React Native OTLP

facostaembrace:
## Goal

- upgrading Embrace iOS to 6.6.0
- fixes covering breaking changes
  - Implementing `EmbraceSemantics`
  - supporting deprecation of `spanBuilder?.markAsKeySpan()`
  - unit tests


packages/core/ios/EmbraceManager.swift
```diff
@@ -482,7 +483,7 @@ class EmbraceManager: NSObject {
                                             ]),
                                             events: eventsFrom(array: []),
                                             // `errorCode` should be used to calc `emb.error_code` attr in native sdk
-                                            errorCode: ErrorCode.failure
+                                            errorCode: SpanErrorCode.failure
```
facostaembrace, line 486: note: breaking change in ios 6.6.0

packages/core/ios/EmbraceManager.swift
```diff
@@ -598,8 +599,6 @@ class EmbraceManager: NSObject {
 
         if !parentSpanId.isEmpty, let parent = spanRepository.get(spanId: parentSpanId) {
             spanBuilder?.setParent(parent)
-        } else {
-            spanBuilder?.markAsKeySpan()
```
facostaembrace, line 602: note: deprecated, `emb.key` is no longer valid and it's not part of the span attributes

packages/react-native-otlp/ios/RNEmbraceOTLP.swift
```diff
@@ -147,7 +151,21 @@ class RNEmbraceOTLP: NSObject {
                         crashReporter = EmbraceCrashReporter()
                     }
 
-                    let servicesBuilder = CaptureServiceBuilder().addDefaults()
+                    let servicesBuilder = CaptureServiceBuilder()
+                    
+                    let urlSessionServiceOptions = URLSessionCaptureService.Options(
+                        // allowing to enable/disable NSF by code
+                        injectTracingHeader: !config.disableNetworkSpanForwarding,
```
facostaembrace, line 158: added the ability of enable/disable NSF even when the full implementation is not part of this PR

packages/core/ios/EmbraceManager.swift
facostaembrace: @ArielDemarco @NachoEmbrace is this cleanup correct? or is there anything else I need to implement to cover this scenario? (given the fact this method is now deprecated). I would like this confirmed before I release this upgrade (planed for today/tomorrow)

facostaembrace: Confirmed by @NachoEmbrace , no action item required. This can be removed completely.

packages/core/ios/EmbraceManager.swift
```diff
@@ -36,6 +38,7 @@ class SDKConfig: NSObject {
         self.disableAutomaticViewCapture = from["disableAutomaticViewCapture"] as? Bool ?? false
         self.endpointBaseUrl = from["endpointBaseUrl"] as? String
         self.disableNetworkSpanForwarding = from["disableNetworkSpanForwarding"] as? Bool ?? false
+        self.ignoredURLs = from.value(forKey: "disabledUrlPatterns") as? [String] ?? []
```
jpmunz, line 41: not sure which is preferred but I see the other lines doing `from["key"]`, could we do the same here or switch the other ones to match this style?

packages/core/ios/EmbraceManager.swift
jpmunz: looks like a couple spots remaining to cleanup on our end:
* packages/core/ios/EmbraceManager.swift:760
* integration-tests/helpers/span.ts:128
* integration-tests/helpers/span.ts:151

packages/core/ios/EmbraceManager.swift
facostaembrace: ah, this is because I pulled that line from `packages/react-native-otlp/ios/RNEmbraceOTLP.swift` which is using `from.value`. will clean this to use `from["xxx"]` but I'll unify the behavior when I work on removing the code duplication

facostaembrace: done

packages/core/ios/EmbraceManager.swift
facostaembrace: done

jpmunz (approved):

PR #310, opened 2024-12-18
[EMBR-6026] Avoid code duplication when starting native SDK (cleanup)

facostaembrace:
## Goal

- iOS: unifying approach to initialize the Embrace.Options object when native sdk starts
- cleanup


packages/core/.gitignore
```diff
@@ -1 +1,3 @@
-src/interfaces/common
\ No newline at end of file
+# Custom
```
jpmunz, line 1: was this meant to say Common?

jpmunz (commented):
Nice, I think this works well as a general strategy for sharing that native code, a few overall questions:

- are you able to edit `packages/common/ios/EmbraceCommon.swift` in Xcode and get proper code completion in the IDE? If you open as part of one of the core or otlp package that would be editing the copy I guess I right and not the original?
- does any of the tooling get confused with `common` being in the `packages` directory but not having a `package.json`? Maybe a quick README would be good to add there to clarify it isn't itself a package
- Is there duplication in Android to tackle as part of this or a separate ticket?

facostaembrace:
> are you able to edit packages/common/ios/EmbraceCommon.swift in Xcode and get proper code completion in the IDE?
Yes, everything is working well in terms of development here

> If you open as part of one of the core or otlp package that would be editing the copy I guess I right and not the original?
This is a fair concern, I can modify the reference to keep this structure but pointing to the file in package/common/ios/EmbraceCommon.swift (now it's pointing the reference to the copied file)


> does any of the tooling get confused with common being in the packages directory but not having a package.json? Maybe a quick README would be good to add there to clarify it isn't itself a package
No as far as I can't tell, but for example I didn't try to run the lerna script to generate a release.
Other scripts are working fine, I think if there is no package.json none of the scripts in the main package.json should affect. But I can take a look

> Is there duplication in Android to tackle as part of this or a separate ticket?
As far as I can recall we have some code duplication in the unit-test suite, maybe I can tackle that as well. In the packages I don't think now there is a good amount of code that could be considered copied/pasted. We have the `start` of the sdk duplicated in both Core and OTLP, but the OTLP is just adding the config of the exporters and that's all. Nothing else is duplicated beside the call to the start method for initializating the native sdk

packages/core/.gitignore
facostaembrace: no, just added # Custom in here because here `packages/react-native-otlp/.gitignore` I saw a bunch of other titles as comment, and then just copied/pasted the same in here. I can remove/tweak if you prefer

jpmunz: ya just didn't seem particularly useful in this case, I think it comes in handy when there's a larger file and we want to put a header on different sections

packages/common/ios/EmbraceCommon.swift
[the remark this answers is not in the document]
facostaembrace: Spoke with @ArielDemarco about it. it doesn't hurt to rename. if for some reason there are two modules with the same name, even when they are different, they could expose those symbols to objective-c and they could collide

jpmunz (approved):

packages/core/.gitignore
facostaembrace: done, removed

facostaembrace:
@jpmunz just for reference for when you are back: moved `common/**` folder out of `packages/` since after all I've decided that it's not technically a package, so it would be better if it is placed outside since it's not going to have neither tests nor package.json file

PR #313, opened 2024-12-20
EMBR-6016 WIP expo config plugin

jpmunz:
WIP for adding an Expo Config plugin


packages/core/src/__tests__/plugin.android.test.ts
```diff
@@ -0,0 +1,295 @@
...
+import {
+  withAndroidEmbraceApplySwazzlerPlugin,
+  withAndroidEmbraceJSONConfig,
+  withAndroidEmbraceOnCreate,
+  withAndroidEmbraceSwazzlerDependency,
+} from "../plugin/withAndroidEmbrace";
+
+import {getMockModConfig, readMockFile} from "./helpers/pluginTestUtils";
+
+// TODO, fails if using `import` here?
```
jpmunz, line 16: @facostaembrace I thought I should be able to do `import * as fs from 'fs';` here but wasn't building unless I did `require`

packages/core/src/plugin/index.ts
```diff
@@ -0,0 +1,21 @@
+import {ConfigPlugin, withPlugins} from "@expo/config-plugins";
+
+import withIosEmbrace from "./withIosEmbrace";
+import withAndroidEmbrace from "./withAndroidEmbrace";
+import {EmbraceProps} from "./types";
+
+const withEmbrace: ConfigPlugin<EmbraceProps> = (config, props) => {
+  // TODO, should be a flag to allow these to be blank for the case that we aren't sending to Embrace
+  if (!(props.iOSAppId && props.androidAppId && props.apiToken)) {
```
facostaembrace, line 9: is there a ticket reflecting this TODO?

packages/core/src/plugin/xcodeproj.ts
```diff
@@ -0,0 +1,123 @@
+/**
+ * This file provides helpers on top of https://github.com/apache/cordova-node-xcode/tree/3.0.1 to perform a handful
+ * of operations not directly supported by that library.
+ *
+ * This is a subset and duplicates some functionality from packages/core/scripts/util/ios.ts, once our automated install
```
facostaembrace, line 5: got it

packages/core/src/__tests__/plugin.android.test.ts
facostaembrace: that's weird. yeah, I saw your comments in other places in this PR and assumed that you didn't find any other way to avoid the problem

packages/core/src/plugin/DEVELOPING.md
```diff
@@ -0,0 +1,21 @@
...
+./update-embrace-packages.sh <app-path>
+./set-embrace-config.js <app-path>  ./embrace-configs/<config>.json --prebuild
+cd <app-path>
+./test-plugin.sh <path>/embrace-react-native-sdk
+```
+Use `../../integration-tests/expo-app-with-prebuild` to test
+the behaviour of the plugin in an actual app.
+
+NOTE: because we end up with `package.json` in `lib/` when building this package the [resolver](https://github.com/expo/expo/blob/814867fd9d0adbd56580eb09be1e81134bb7466e/packages/%40expo/config-plugins/src/utils/plugin-resolver.ts)
+used by @expo/config-plugins gets confused and thinks that is the module root so we need to make sure `app.plugin.js`
```
facostaembrace, line 20: got it

facostaembrace (approved):
This is great @jpmunz! .
I'm thinking about the next steps for this feature. I believe for the OTLP package we have a lot of code written in the README.md file for users that want to initialize Embrace in the native layer without the `appId` and adding the Grafana tokens + apis (+ the tweaks each platform initialization needs for sending data outside Embrace). This plugin can be extended to cover that case as well

PR #314, opened 2024-12-26
[EMBR-4174] Allow / Add instructions for running the RN SDK w/o an Embrace appId or token

facostaembrace:
## Goal

- Allowing iOS to initialize SDK without appId / token
- Android doesn't require any change beside what docs already mention


common/interfaces/index.ts
```diff
@@ -3,28 +3,19 @@
...
   disableNetworkSpanForwarding?: boolean;
   endpointBaseUrl?: string;
+  disabledUrlPatterns?: string[];
 }
 
 // Today Android is not configurable through code,
 // this is a placeholder for future implementations.
 interface AndroidConfig {}
 
-interface IOSConfig {
```
facostaembrace, line 18: duplicated, removing

common/ios/RNEmbraceCommon.swift
```diff
@@ -59,9 +59,18 @@ func initEmbraceOptions(config: SDKConfig, exporters: OpenTelemetryExport?) -> E
                                           developmentBaseURL: config.endpointBaseUrl!,
                                           configBaseURL: config.endpointBaseUrl!)
         }
-        
+                
+        if (config.appId == nil && exporters != nil) {
```
jpmunz, line 63: could you explain this additional check? Would something be incorrect if the `.init` block below was just executed instead in the case that there are exporters but no app id?

packages/core/src/index.ts
```diff
@@ -82,33 +82,47 @@ const initialize = async ({
 }: {patch?: string; sdkConfig?: SDKConfig} = {}): Promise<boolean> => {
   const hasNativeSDKStarted = await EmbraceManagerModule.isStarted();
 
+  // if the sdk started in the native side the follow condition doesn't take any effect.
+  // neither iOS setup() nor start() will be overridden
```
jpmunz, line 86:
```suggestion
  // if the sdk has already been started in the native side then the follow condition doesn't have any effect.
  // neither iOS setup() nor start() will be overridden
```

packages/core/src/index.ts
```diff
@@ -82,33 +82,47 @@ const initialize = async ({
...
 
     let isStarted;
     try {
-      isStarted = customStartEmbraceSDK
-        ? await customStartEmbraceSDK(startSdkConfig)
-        : await EmbraceManagerModule.startNativeEmbraceSDK(startSdkConfig);
+      if (startCustomExport) {
+        if (!startSdkConfig.appId) {
+          console.log(
+            "[Embrace] 'sdkConfig.ios.appId' not found, custom exporters will be used",
```
jpmunz, line 110:
```suggestion
            "[Embrace] 'sdkConfig.ios.appId' not found, only custom exporters will be used",
```

common/ios/RNEmbraceCommon.swift
facostaembrace: basically the appId can't be `nil` using that signature. it doesn't compile
image

jpmunz: isn't that case handled by the fallback value here `appId: config.appId ?? "",`?

facostaembrace: no, that compiles but it doesn't initialize the SDK, it throw an error and the native side rejects

jpmunz (approved):

PR #315, opened 2024-12-27
[EMBR-5093] Add more react friendly interface for starting the embrace SDK in the core package

facostaembrace:
## Goal

- require.context reference
  - https://github.com/facebook/metro/issues/666#issuecomment-1742186952
  - https://medium.com/@dannyhyunsoowilliams/dynamic-imports-supported-in-react-native-e839f6481eb2
  - https://github.com/facebook/metro/issues/52#issuecomment-2265175282


packages/core/src/index.ts
```diff
@@ -77,43 +80,62 @@ const isObjectNonEmpty = (obj?: object): boolean =>
   Object.keys(obj || {}).length > 0;
 
 const initialize = async ({
-  patch,
   sdkConfig,
+  patch,
 }: {patch?: string; sdkConfig?: SDKConfig} = {}): Promise<boolean> => {
+  const logger = new EmbraceLogger(console, sdkConfig?.debug ?? true);
```
jpmunz, line 86: for the `debug` flag I'd still expect the logger to output warning + error level messages but it seems that this turns off all logging. I think we should either have this boolean debug log messages or instead change the parameter from a boolean to a log level enum

packages/core/src/utils/otlp.ts
```diff
@@ -0,0 +1,59 @@
...
+
+class EmbraceOTLP {
+  public module: OTLPModule | null = null;
+  private logger: EmbraceLogger;
+
+  constructor(logger: EmbraceLogger) {
+    this.logger = logger;
+  }
+
+  public get = () => {
```
jpmunz, line 25: alternatively would it be enough to see if `NativeModules.RNEmbraceOTLP` is undefined to check whether the package is available?

packages/core/src/hooks/useEmbrace.ts
```diff
@@ -0,0 +1,43 @@
...
+      };
+
+      try {
+        const isEmbraceStarted = await initialize(config);
+
+        setIsPending(false);
+        setIsStarted(isEmbraceStarted);
+      } catch (error) {
+        setIsPending(false);
+      }
```
jpmunz, line 29: could use `finally` here instead since it seems we'll always want pending turned off:

```suggestion
      try {
        const isEmbraceStarted = await initialize(config);
        setIsStarted(isEmbraceStarted);
      } finally {
        setIsPending(false);
      }
```

packages/core/src/index.ts
facostaembrace: I like the idea of turning the boolean into a log level enum, I'll take that one

packages/core/src/utils/otlp.ts
facostaembrace: no, this approach doesn't work. I just took a moment to verify it again and I'm able to confirm that even when `NativeModules.RNEmbraceOTLP` would be `null` if the package is not installed, it will throw anyway because Metro processes this during the bundling phase

facostaembrace: Screenshot 2025-01-08 at 2 49 25 PM

jpmunz:
@facostaembrace as part of this could you also update the code snippets in these spots to reflect using the new helper?
* packages/react-native-tracer-provider/README.md
* packages/react-native-otlp/README.md

packages/core/src/utils/EmbraceLogger.ts
facostaembrace, line 1: @jpmunz fyi I've renamed this file and moved the old `logger.ts` into `utils`. let me know if this change make sense for you

packages/core/src/index.ts
```diff
@@ -2,74 +2,93 @@
...
+  sdkConfig?: SDKConfig;
+  debug?: boolean;
+}
+
+const initialize = async (
+  {sdkConfig, patch, debug}: EmbraceInitArgs = {debug: true},
+): Promise<boolean> => {
+  const logger = new EmbraceLogger(
+    console,
+    debug ? ["info", "warn", "error"] : ["warn", "error"],
```
jpmunz, line 25: minor, I was picturing something a bit different here where the parameter would be a single `logLevel` and setting it to "info", etc. would then log anything at that level or above

if we don't change that I think at least renaming `debug` since that seems to imply something extra to set during development but we have it on by default it

packages/react-native-tracer-provider/README.md
```diff
@@ -38,26 +38,12 @@ the SDK before the component that sets up the provider renders, or you can make
 `useEmbraceNativeTracerProvider` hook to prevent it from triggering until the SDK is ready as in this example:
 
 ```javascript
-  const [embraceSDKLoaded, setEmbraceSDKLoaded] = useState<boolean>(false);
-  const {tracerProvider} = useEmbraceNativeTracerProvider({}, embraceSDKLoaded);
+  const {isStarted} = useEmbrace({
+    ios: SDK_CONFIG,
+    debug: true,
```
jpmunz, line 43: I'd maybe leave off `debug: true` to keep the example minimal

packages/react-native-otlp/README.md
```diff
@@ -36,66 +36,47 @@ For this example we will use Grafana Cloud in terms of redirecting telemetry dat
...
-            {
-              key: "Authorization",
-              token: `Basic ${token}`,
-            },
-          ],
-        },
-      }),
-    [],
-  );
+  const {isPending} = useEmbrace({
```
jpmunz, line 73: in `packages/react-native-tracer-provider/README.md` it works off `isStarted`, should we keep it consistent here? Is there another reason to use `isPending`?

packages/react-native-otlp/README.md
jpmunz, line 1: not related to the changes here but when I opened the markdown I noticed this other link is rendered incorrectly:
image

packages/react-native-otlp/README.md
```diff
@@ -36,66 +36,47 @@ For this example we will use Grafana Cloud in terms of redirecting telemetry dat
...
+      {
+        key: "Authorization",
+        token: `Basic ${GRAFANA_TOKEN}`,
+      },
+    ],
+  },
+};
+
+// this is the minimum of configuration needed at this point,
+// for more information please refer to docs under `@embrace-io/react-native`.
```
jpmunz, line 68: I'd add a quick relative link if possible, I didn't test but I think this might work:

```suggestion
// for more information please refer to docs under [`@embrace-io/react-native`](../core).
```

integration-tests/test-harness/src/EmbraceTestHarness.tsx
```diff
@@ -1,45 +1,29 @@
...
+  if (!allowCustomExport) {
+    sdkConfig.exporters = undefined;
+  }
 
-    init();
-  }, [allowCustomExport]);
+  const {isPending} = useEmbrace(sdkConfig);
 
-  if (!embraceLoaded) {
+  if (isPending) {
```
jpmunz, line 26: similar comment here, though I guess `isPending` does make sense for showing the loading screen...maybe we need a 2nd block then for `!(isPending || isStarted)` to indicate an error state

packages/core/src/index.ts
```diff
@@ -2,74 +2,93 @@
...
 
       return Promise.resolve(false);
     }
 
-    const {startCustomExport: customStartEmbraceSDK, ...originalSdkConfig} =
-      sdkConfig || {};
+    const {exporters: otlpExporters, ...originalSdkConfig} = sdkConfig || {};
+    const embraceOTLP = new EmbraceOTLP(logger);
+
+    try {
```
jpmunz, line 41: is this earlier check required? Since only `set` later on depends on it being called I'm wondering if this could be removed and we collapse everything to a single call later on similar to what we had previously, e.g.:
```
      isStarted = otlpExporters
        ? await otlpStart(logger, otlpExpoters)
        : await EmbraceManagerModule.startNativeEmbraceSDK(startSdkConfig);
```

packages/core/src/utils/EmbraceOTLP.ts
```diff
@@ -0,0 +1,56 @@
...
+
+import EmbraceLogger from "./EmbraceLogger";
+
+interface Package {
+  initialize: (
+    otlpExporterConfig: OTLPExporterConfig,
+  ) => (sdkConfig: IOSConfig | AndroidConfig) => Promise<boolean>;
+}
+
+const RN_EMBRACE_OTLP_PATH =
```
jpmunz, line 11: could you add a comment here with an example of what the current working directory would be? Are the extra ".." required because we need to move backwards out of where the core package is installed in node modules?

packages/core/src/index.ts
facostaembrace: > minor, I was picturing something a bit different here where the parameter would be a single logLevel and setting it to "info", etc. would then log anything at that level or above

but if we goes with this way we would agree on having

type Level = 'info' | 'warn' | 'error';

error level -> will contain logs for error/warn/info (everything)
warn level -> will contain logs for warn/info
info level -> will contain logs only for info

and you would never be able to disable info logs. if that's what we want then I can go with that, I was just thinking on the user having the ability of disabling everything

jpmunz: would be the other way, so:

error level -> will contain logs only for error
warn level -> will contain logs for warn/error
info level -> will contain logs only for error/warn/info (everything)

so then you'd always get errors and then you can get more levels as you increase towards info

packages/react-native-otlp/README.md
facostaembrace: no other reason beside the fact that the example is showing a "loading indicator", maybe this one could also use the `isStarted` boolean flag adding the 2nd block indicating the error state as you pointed in other comment here (and do the same in `packages/react-native-tracer-provider/README.md`)

packages/core/src/index.ts
facostaembrace: okok, got it. will work on it

packages/react-native-otlp/README.md
jpmunz: ah right

packages/core/src/utils/EmbraceOTLP.ts
facostaembrace: yeah, the extra ".." are because we should point to the exact place where we expect to find the OTLP index.js file.
The structure we expect is kind of the folllowing: 

```
node_modules
└── @embrace-io
    ├── react-native
    │   └── lib
    │       └── src
    │           └── utils
    │               ├── EmbraceOTLP.js
    │               └── ...
    └── react-native-otlp
        └── lib
            └── src
                ├── index.js
                └── ...
```

facostaembrace: added a comment into the code as well for reference

packages/core/src/index.ts
facostaembrace: refactored as discussed earlier

integration-tests/test-harness/package.json
```diff
@@ -12,10 +12,10 @@
   "main": "lib/index.js",
   "typings": "lib/index.d.ts",
   "peerDependencies": {
-    "@embrace-io/react-native": "*",
-    "@embrace-io/react-native-otlp": "*",
-    "@embrace-io/react-native-tracer-provider": "*",
-    "@opentelemetry/instrumentation-react-native-navigation": "*",
+    "@embrace-io/react-native": "file:../artifacts/embrace-io-react-native-local.tgz",
```
jpmunz, line 15: was there a reason for switching these? The current setup with "*" was to allow the test apps to swap out these dependencies for different versions, e.g. maybe an already released one from npm instead of the local artifact

packages/core/src/index.ts
```diff
@@ -2,92 +2,92 @@
...
-const tracking = require("promise/setimmediate/rejection-tracking");
+interface EmbraceInitArgs {
+  patch?: string;
+  sdkConfig?: SDKConfig;
+  debug?: EmbraceLoggerLevel;
+}
 
-const UNHANDLED_PROMISE_REJECTION_PREFIX = "Unhandled promise rejection";
+const initialize = async (
+  {sdkConfig, patch, debug}: EmbraceInitArgs = {debug: "info"},
```
jpmunz, line 21: can we rename `debug` to `logLevel`?

packages/core/src/index.ts
```diff
@@ -2,92 +2,92 @@
...
+          logger.log(
+            "'sdkConfig.ios.appId' not found, only custom exporters will be used",
           );
         }
 
-        isStarted = await startCustomExport(startSdkConfig);
-      } else {
-        isStarted =
-          await EmbraceManagerModule.startNativeEmbraceSDK(startSdkConfig);
+        const otlp = new EmbraceOTLP(logger);
```
jpmunz, line 53: is there a need for a class any longer? Could it be `oltpGetStart(logger, otlpExporters)` instead?

packages/react-native-tracer-provider/README.md
```diff
@@ -28,27 +28,14 @@ renders, or you can make use of the `enabled` parameter on the `useEmbraceNative
...
-        setEmbraceSDKLoaded(true);
-      }
-    };
-
-    init();
-  }, []);
+const optionalConfig = {...}; // See EmbraceNativeTracerProviderConfig in ./src/types/ for possible options
+const {tracerProvider: embraceTracerProvider, tracer: embraceTracer} = useEmbraceNativeTracerProvider(optionalConfig, isStarted);
+
+const {tracerProvider} = useEmbraceNativeTracerProvider({}, isStarted);
```
jpmunz, line 38: I think this line got added as a duplicate:
```suggestion
```

packages/react-native-otlp/README.md
```diff
@@ -31,76 +31,61 @@ cd ios && pod install --repo-update
...
-            },
-          ],
-        },
-      }),
-    [],
-  );
+  const {isPending, isStarted} = useEmbrace({
+    ios: SDK_CONFIG,
+    exporters: EXPORT_CONFIG,
+    debug: true,
```
jpmunz, line 76:
```suggestion
```

integration-tests/test-harness/package.json
facostaembrace: no, actually I thought I've reverted this change. I'll take a look

facostaembrace: ah, I see. I've reverted that change here https://github.com/embrace-io/embrace-react-native-sdk/pull/317/commits/4f10babd64aebbe60d08c4ea08868243c6fbf12a but not in this PR. cleaning up as well

packages/core/src/index.ts
facostaembrace: I would create the helper acting as a wrapper but I wouldn't refactor the entire logic in EmbraceOTLP.ts file since that would require a new pass of cleaning up/testing. would that be enough?

packages/core/src/hooks/useEmbrace.ts
```diff
@@ -0,0 +1,46 @@
+import {useEffect, useState} from "react";
+
+import {SDKConfig, EmbraceLoggerLevel} from "../interfaces";
+
+import {initialize} from "./../index";
```
jpmunz, line 5:
```suggestion
import {initialize} from "../index";
```

package.json
```diff
@@ -18,7 +18,7 @@
     "lint:clang:fix": "./scripts/run-clang-format.sh --fix",
     "lint:swift": "./scripts/run-swiftlint.sh",
     "lint:swift:fix": "./scripts/run-swiftlint.sh --fix",
-    "test": "jest",
+    "test": "jest --maxWorkers=2",
```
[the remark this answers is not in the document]
facostaembrace: @jpmunz added a KTLO task to revisit this error https://www.notion.so/embraceio/Revisit-JS-unit-tests-for-cleaning-up-console-logs-avoiding-tests-fail-silently-17a7e3c9985280129cb5d21e407b2b0d since I was able to see it again randomly even after this change

common/js/interfaces.ts
```diff
@@ -2,6 +2,12 @@
  * Common interfaces shared by Embrace Packages
  */
 
+interface SDKConfig {
+  ios?: IOSConfig;
+  exporters?: OTLPExporterConfig;
+  debug?: boolean;
```
jpmunz, line 8: is this still needed now that we have `logLevel` as a param to initialize?

jpmunz (approved):

PR #317, opened 2025-01-02
[EMBR-5914] Remove old packages from RN SDK

facostaembrace:
## Goal

- https://www.notion.so/embraceio/Remove-embrace-screen-orientation-package-10c7e3c99852809d9063d209f9a0b4b3 | removing `@embrace-io/react-native-orientation-change-tracker`, refactoring hook and moving it into `core`.
- https://www.notion.so/embraceio/Remove-webview-tracker-10c7e3c99852806bbbe8ed0a58bfc834 | removing `@embrace-io/webview-tracker`. Not replacing.
- https://www.notion.so/embraceio/Removal-of-all-interfaces-providers-helpers-dealing-w-Axios-1627e3c998528016b82ac928977b77e4 | removing all related to `axios`. Not replacing.


packages/core/src/hooks/useOrientationListener.ts
```diff
@@ -0,0 +1,101 @@
...
+    },
+    [],
+  );
+
+  /**
+   * Initialize the listener
+   */
+  useEffect(() => {
+    if (!initRef.current) {
+      const init = () => {
```
jpmunz, line 83: what is the reason for wrapping this section in the `init` function?

packages/core/src/hooks/useOrientationListener.ts
```diff
@@ -0,0 +1,101 @@
...
+        if (orientationRef.current) {
+          logOrientationChange(orientationRef.current);
+        }
+
+        Dimensions.addEventListener("change", handleOrientationChange);
+      };
+
+      init();
+
+      // initialized just once
```
jpmunz, line 95: could you guarantee this instead by just setting `[]` for the dependency on the `useEffect`?

packages/core/src/hooks/useOrientationListener.ts
```diff
@@ -0,0 +1,101 @@
+import {Dimensions, ScaledSize} from "react-native";
+import {useCallback, useEffect, useRef} from "react";
+
+import EmbraceLogger from "../logger";
+import {addBreadcrumb} from "..";
+
+type ScreenOrientation = "portrait" | "landscape";
+type Orientation = ScreenOrientation | undefined;
```
jpmunz, line 8: is this additional type necessary? Could `ScreenOrientation?` be used in places where it is optional, or maybe instead definition it as `type ScreenOrientation = "portrait" | "landscape" | undefined;` ?

packages/core/src/hooks/useOrientationListener.ts
facostaembrace: I think there is no reason actually. this is is not an async operation so the wrapping the logic into another callback is not giving extra benefits. removing, good call out

packages/core/src/hooks/useOrientationListener.ts
facostaembrace: I made couple of changes improving this, I've moved the callback directly into the listener so it's not needed the dependency

packages/core/src/hooks/useOrientationListener.ts
facostaembrace: removed `Orientation`, and used `undefined` on demand when this was needed

jpmunz (approved):

PR #318, opened 2025-01-02
[EMBR-5087] Remove embrace apollo-graphql package

facostaembrace:
## Goal

- NOT REMOVING YET, working on a POC using our @embrace-io/react-native-apollo-graphql package for learning about how this should work and what it does.
- Conclusions at the moment: it doesn't work.


jpmunz:
will remove the package for now since it isn't working correctly, can revisit if we have need for it in the future

PR #321, opened 2025-01-07
[EMBR-5409] Add an app template for rn76

facostaembrace:
## Goal

- added new template
- upgraded @react-native-community/cli package due to local issues regarding v0.76
- adding rn76 app to browserstack


integration-tests/templates/rn76-template/package.json
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "ProjectName",
```
jpmunz, line 2: not exactly sure if this matters but to make it consistent with the other ones:
```suggestion
  "name": "rn76-template",
```

jpmunz (approved):

integration-tests/templates/rn76-template/package.json
facostaembrace: it doesn't matter because when you use the template and put a name there that one is used instead of what add this package.json file. I'll push the change just for consistency but it doesn't take any effect

PR #324, opened 2025-01-08
EMBR-5089 replace spans package

jpmunz:
Removing most of the functionality and adding a `startView` convenience method to the tracer provider package


jpmunz:
@facostaembrace the JS tests will fail on this for now because the action-tracker + react-native-navigation + react-navigation packages still reference the methods I removed, can wait to merge this until after those packages have been removed/updated

packages/core/android/src/main/java/io/embrace/rnembracecore/EmbraceManagerModule.java
```diff
@@ -465,127 +448,6 @@ public void logNetworkClientError(String url,
 
     }
 
-    @ReactMethod()
-    public void startSpan(String name, String parentSpanId, Double startTimeMs, Promise promise) {
```
facostaembrace, line 469: nice cleanup!

facostaembrace (approved):

PR #325, opened 2025-01-10
EMBR-3956 removing logScreen, setUserAsPayer, clearUserAsPayer, logRNAction

jpmunz:


packages/core/android/src/main/java/io/embrace/rnembracecore/EmbraceManagerModule.java
```diff
@@ -253,21 +253,6 @@ public void setReactNativeVersion(String version, Promise promise) {
         }
     }
 
-    @ReactMethod
-    public void logRNAction(String name, Double startTime, Double endTime, ReadableMap properties, Integer payloadSize, String output, Promise promise) {
```
jpmunz, line 257: this one was already removed on the JS layer

facostaembrace (approved):

PR #326, opened 2025-01-10
EMBR-6220 add convenience methods for the tracer provider

jpmunz:


packages/react-native-tracer-provider/src/convenience.ts
facostaembrace, line 1: for the core package I started to export everything at the bottom of the file instead of exporting every method inline. should we agree on one way of doing it for consistency?

packages/react-native-tracer-provider/src/convenience.ts
```diff
@@ -10,3 +20,63 @@ export const startView = (tracer: Tracer, viewName: string): Span =>
       "emb.type": "ux.view",
     },
   });
+
+/**
+ * endAsFailed offers a shortcut for setting the span's status as ERROR and then ending it
+ */
+export const endAsFailed = (span: Span) => {
```
facostaembrace, line 27: for the core package I started to export everything at the bottom of the file instead of exporting every method inline. should we agree on one way of doing it for consistency?

PR #327, opened 2025-01-13
fix warnings

jpmunz:


integration-tests/test-harness/src/EmbraceTestHarness.tsx
```diff
@@ -3,7 +3,7 @@ import * as React from "react";
 import {initialize as initEmbrace} from "@embrace-io/react-native";
 import {Text, View} from "react-native";
 import {styles} from "./helpers/styles";
-import {SDKConfig} from "@embrace-io/react-native/lib/src/interfaces/Config";
+import {SDKConfig} from "@embrace-io/react-native/lib/src/interfaces/common/interfaces";
```
jpmunz, line 6: should we be re-exporting `SDKConfig` from the core package?

packages/react-native-tracer-provider/src/useEmbraceNativeTracerProvider.ts
```diff
@@ -71,6 +71,7 @@ export const useEmbraceNativeTracerProvider = (
         })
         .finally(() => setIsLoading(false));
     }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
jpmunz, line 74: @facostaembrace how do you feel about just ignoring the rule here instead? The above logic should be safe to re-run if `tracerProvider` changes value but it wouldn't be necessary to do so

facostaembrace: tbh I don't like the practice of ignoring dependencies or putting this kind of rules off. usually there are ways of refactoring the code in a way the dependency is really not needed.
this often leads to unexpected problems or side effects, but I think it's just fine by now. we can revisit in the future in case we see something weird around it

facostaembrace (approved):

packages/react-native-tracer-provider/src/useEmbraceNativeTracerProvider.ts
jpmunz: np that makes sense, I think the refactor that would remove the warning in this case is removing the `tracerProvider` from the else check altogether like:

```
if (!started) {
  setError(
    "The Embrace SDK must be started to use the TracerProvider, please invoke `initialize` from `@embrace-io/react-native`.",
  );
  setIsError(true);
} else if (!tracerProvider) {
  const provider = new EmbraceNativeTracerProvider(config);
  setTracerProvider(provider);
  setTracer(provider.getTracer("embrace-default-tracer"));
}
```

I think that's safe I'm just a little hesitant to remove the guard in case there's some way I'm not thinking of that the block could execute again and create a 2nd tracer provider. I think between that option and just adding it to the dependency list and having a potential re-render that doesn't do anything I prefer the 2nd, let me know what you think

PR #332, opened 2025-01-16
(fix) refactored JS unit tests for `index.ts` file

facostaembrace:
## Goal

- refactored tests that cover the `initialize` main method for RN sdk
- console log/warn messages displayed across the workflow silenced at least for this file for now and adding checks around those (those were making the workflow noisy and difficult to inspect in case real failures) .

### DIFF
--- a/packages/core/src/__tests__/index.test.ts
+++ b/packages/core/src/__tests__/index.test.ts
@@ -1,26 +1,31 @@
 import {waitFor} from "@testing-library/react-native";
 
 import {oltpGetStart} from "../utils/otlp";
-import {initialize} from "../index";
+import {AndroidConfig, initialize, IOSConfig} from "../index";
 import {handleGlobalError} from "../api/error";
 import {ComponentError, logIfComponentError} from "../api/component";
 
-const testValue = "Value";
-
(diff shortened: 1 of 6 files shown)


packages/core/src/__tests__/index.test.ts
```diff
@@ -58,73 +64,70 @@ jest.mock("../EmbraceManagerModule", () => ({
...
 
-    expect(result).toBe(true);
-    expect(mockSetReactNativeVersion).toHaveBeenCalledWith("0.75.4");
-    expect(mockSetJavaScriptPatchNumber).toHaveBeenCalledWith(testValue);
-    expect(mockSetReactNativeSDKVersion).toHaveBeenCalledWith(
-      expect.any(String),
-    );
-    expect(mockStart).not.toHaveBeenCalled();
-    expect(mockLogMessageWithSeverityAndProperties).toHaveBeenCalled();
-    expect(mockLogMessageWithSeverityAndProperties.mock.calls[0][0]).toBe(
```
jpmunz, line 99: as discussed we should still assert that the enable tracking helper was called but can verify the specific functionality of what it's doing in error.test.ts

jpmunz (approved):

PR #333, opened 2025-01-17
EMBR-5086 update actions package

jpmunz:


integration-tests/test-harness/src/screens/ReduxTestingScreen/reducers.ts
```diff
@@ -0,0 +1,33 @@
+import {combineReducers, Reducer} from "@reduxjs/toolkit";
+import {initialCounterState, CounterState, CounterActions} from "./actions";
```
facostaembrace, line 2: are we duplicating the content of this file? I see the same under `packages/react-native-redux/__tests__/helpers/store.ts`. is there a reason why we would need it twice?

facostaembrace (approved):

PR #346, opened 2025-01-20
EMBR-6231 update ios to 6.7.0

jpmunz:
Did a basic check that the app loaded with the updated version, will save more testing for 6.0 QA


.github/workflows/ios.unit-test.yml
```diff
@@ -42,6 +42,7 @@ jobs:
             - 'packages/**/*.swift'
             - 'packages/**/*.m'
             - 'packages/**/*.h'
+            - 'packages/**/test-project/ios/Podfile.lock'
```
jpmunz, line 45: wanted the iOS tests to run against the new version but realized they weren't because none of the above file types were modified. Not sure if there's a better thing to point to when bumping the version, I did have to remember to update this manually in each packages 'test-project'

facostaembrace (approved):

PR #357, opened 2025-01-21
(fix) Upgrading basic-test-app

facostaembrace:
## Goal

- upgrading basic-test-app, matching react-native version with workspace


jpmunz (approved):
this looks good to me, I'm curious if you tried switching the install strategy to linked and if that allowed it to continue to work with 0.74.1? I think something to keep in mind next time we have an issue with this app running

facostaembrace:
Not merging yet since it seems like there are still issues. Having this PR in place help the basic-test-app to compile but it fails at runtime.

PR #373, opened 2025-01-28
EMBR-6294 useEmbraceIsStarted hook, extra logging, warn about multiple initialize calls

jpmunz:


packages/core/README.md
```diff
@@ -52,40 +52,89 @@ To run these steps manually refer to [this section of our guide](https://embrace
 
 ## Initialize the Embrace SDK
 
-Initialize method applies the necessary listener to your application. This allow Embrace to track javascript errors, check js bundle changes (if you use OTA), track js patch and react native versions.
```
jpmunz, line 55: copied from the docs branch

packages/core/src/hooks/useEmbraceIsStarted.test.ts
```diff
@@ -0,0 +1,50 @@
...
+}));
+
+describe("useEmbraceIsStarted", () => {
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it("should return if the Native SDK is started", async () => {
+    mockIsStarted.mockResolvedValue(true);
+    const {result} = renderHook(() => useEmbraceIsStarted());
```
facostaembrace, line 20:
```suggestion
    const {result} = renderHook(useEmbraceIsStarted);
```

packages/core/src/hooks/useEmbraceIsStarted.test.ts
```diff
@@ -0,0 +1,50 @@
...
+    expect(result.current.isChecking).toBe(true);
+    await waitFor(() => {
+      expect(result.current.isChecking).toBe(false);
+      expect(result.current.isStarted).toBe(true);
+    });
+  });
+
+  it("should return if the Native SDK is not started", async () => {
+    mockIsStarted.mockResolvedValue(false);
+    const {result} = renderHook(() => useEmbraceIsStarted());
```
facostaembrace, line 31:
```suggestion
    const {result} = renderHook(useEmbraceIsStarted);
```

packages/core/src/hooks/useEmbraceIsStarted.test.ts
```diff
@@ -0,0 +1,50 @@
...
+    expect(result.current.isChecking).toBe(true);
+    await waitFor(() => {
+      expect(result.current.isChecking).toBe(false);
+      expect(result.current.isStarted).toBe(false);
+    });
+  });
+
+  it("should handle getting an error when checking if the Native SDK is started", async () => {
+    mockIsStarted.mockRejectedValue("some error");
+    const {result} = renderHook(() => useEmbraceIsStarted());
```
facostaembrace, line 42:
```suggestion
    const {result} = renderHook(useEmbraceIsStarted);
```

packages/core/src/hooks/useEmbraceIsStarted.ts
```diff
@@ -0,0 +1,26 @@
+import {useEffect, useState} from "react";
+
+import {EmbraceManagerModule} from "../EmbraceManagerModule";
+
+const useEmbraceIsStarted = () => {
+  const [isChecking, setIsChecking] = useState(true);
```
facostaembrace, line 6: Is there a reason why we don't call this flag like in the other hook? `isPending` (or maybe `isLoading`). just for consistency I guess it would be nice.
In any case, I would add some comments here explaining what means `isChecking` or under which situation it would be useful for the user to pull it and use

facostaembrace (approved):

packages/core/src/hooks/useEmbraceIsStarted.ts
jpmunz: ya I was struggling with this one, I was worried that `isPending` or `isLoading` could be interpreted as the SDK itself being pending or loading whereas actually it's the hook that hasn't finished checking yet what state the SDK is in, maybe a couple other options if one seems less confusing:
* have a single `isStarted` variable returned that can be `true`, `false`, or `null`, if it's `null` it means we haven't determined yet if the SDK is started or not
* rename to something like `useEmbraceGetState` and return `state` which is an enum that could be "started", "not-started", "checking"/"unknown"
* rename to something like `useEmbraceCheckIsStarted` which maybe then makes the use of the `isChecking` value more obvious

facostaembrace: I think `null`, `true` or `false` make more sense to me at this point + a comment explaining that the `null` value could be found (since it's waiting for an async operation to come back with the real information)

facostaembrace: Also I was considering that this hook could receive a boolean flag that enables the hook to re-run in case for some reason a component needs to re-check this information to provide more value. What do you think about it?

Having no dependencies means this hook will just run when the component that call it is mounted, so if for some reason users want to control when they want to verify the start of the SDK they wouldn't be allowed.
Kind of similar on how `react-query` works, like `useQuery` receives a `enabled` param that allows to control when the hook should call the desired API.

jpmunz: hmm I guess it comes down to how it is used but the only case I can think of where they would do something prior that would affect the result would be calling `useEmbrace` which would itself return that information

PR #375, opened 2025-01-29
EMBR-5980 fix sourcemap upload

jpmunz:
Previously we triggered source map generation on iOS by updating the react native build phase with:
```
export SOURCEMAP_FILE="$CONFIGURATION_BUILD_DIR/main.jsbundle.map"
```

when this variable is set the react native xcode script emits a map file:
https://github.com/facebook/react-native/blob/17c34295a18c707ea28dc7f6f8566bdb57472b83/packages/react-native/scripts/react-native-xcode.sh#L127

however the problem was with our naming, "$CONFIGURATION_BUILD_DIR/main.jsbundle.map" is exactly the same name that gets used for `PACKAGER_SOURCEMAP_FILE`: https://github.com/facebook/react-native/blob/17c34295a18c707ea28dc7f6f8566bdb57472b83/packages/react-native/scripts/react-native-xcode.sh#L133 and the last thing the script does is `rm $PACKAGER_SOURCEMAP_FILE` which deletes the source map file we just generated: https://github.com/facebook/react-native/blob/17c34295a18c707ea28dc7f6f8566bdb57472b83/packages/react-native/scripts/react-native-xcode.sh#L184

even with this problem sourcemap upload continued to work for some apps because the upload script we package with the pod attempts to create the sourcemap if it isn't able to find it: https://github.com/embrace-io/go/blob/master/tool/ios-upload/run.sh#L114. This assumed the entry file was "index.js" which worked for bare RN apps but failed for Expo apps that had a different entry point

just changing the filename isn't enough because `basename "$SOURCEMAP_FILE"` is used to get `PACKAGER_SOURCEMAP_FILE`'s name so it'll end up with the same problem. Solution here is to create a new directory in the build folder and place our sourcemap in there which protects it from getting removed

also updating manual instructions in docs: https://github.com/embrace-io/embrace-docs/pull/182

### DIFF
--- a/packages/core/scripts/__tests__/__mocks__/ios/testMock.xcodeproj/project.pbxproj
+++ b/packages/core/scripts/__tests__/__mocks__/ios/testMock.xcodeproj/project.pbxproj
@@ -286,7 +286,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "set -e\n\nWITH_ENVIRONMENT=\"../node_modules/react-native/scripts/xcode/with-environment.sh\"\nexport SOURCEMAP_FILE=\"$CONFIGURATION_BUILD_DIR/main.jsbundle.map\";\nREACT_NATIVE_XCODE=\"../node_modules/react-native/scripts/react-native-xcode.sh\"\n\n/bin/sh -c \"$WITH_ENVIRONMENT $REACT_NATIVE_XCODE\"\n";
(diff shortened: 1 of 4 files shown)


facostaembrace (approved):

facostaembrace:
should we mention this fix/change in the "migration to 6.x" section for people that already have the build phases already in place or already ran the install scripts for previous versions?

jpmunz:
> @jpmunz should we mention this fix/change in the "migration to 6.x" section for people that already have the build phases already in place or already ran the install scripts for previous versions?

ya I think that's reasonable, I'll add the note in https://github.com/embrace-io/embrace-docs/pull/182

PR #400, opened 2025-02-18
[EMBR-6585] Read upload token from env var

facostaembrace:
## Goal

- reading from local env variables and removing sensitive values from the repo.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -97,3 +97,5 @@ nx.json
 
 # Yarn
 .yarn
+
+.env
--- a/integration-tests/set-embrace-config.js
+++ b/integration-tests/set-embrace-config.js
@@ -53,6 +53,7 @@
   const fs = require("fs");
   const {Command} = require("commander");
   const program = new Command();
+  const loadEnvVar = require("../config/env");
   let appPath, configPath;
 
   program
@@ -81,7 +82,9 @@
       endpoint: string;
     }
   */
+
   const config = JSON.parse(fs.readFileSync(configPath, "utf8"));
+
   if (options.namespace && config.endpoint) {
     config.endpoint = config.endpoint.replace("<namespace>", options.namespace);
   }
@@ -109,7 +112,7 @@
   const androidConfigPath = `${appPath}/android/app/src/main/embrace-config.json`;
   const androidConfig = {
     app_id: config.android_app_id,
-    api_token: config.api_token,
+    api_token: loadEnvVar("EMBRACE_ORG_TOKEN", configPath),
     sdk_config: {
       app_framework: "react_native",
     },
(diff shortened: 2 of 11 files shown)


facostaembrace:
@jpmunz I think the same can be done without using an extra package as dependency but just for quickness I introduced the dotenvx library to handle local variables using different env files. please, let me know if you have concerns about it

package.json
```diff
@@ -55,5 +55,8 @@
     "react-test-renderer": "^19.0.0",
     "ts-jest": "^29.2.5",
     "typescript": "^5.6.3"
+  },
+  "dependencies": {
+    "@dotenvx/dotenvx": "^1.36.0"
```
jpmunz, line 60: I guess doesn't matter too much since this is for the mono-repo itself, but should it be under `devDepedencies` instead?

integration-tests/set-embrace-config.js
```diff
@@ -109,7 +112,7 @@
   const androidConfigPath = `${appPath}/android/app/src/main/embrace-config.json`;
   const androidConfig = {
     app_id: config.android_app_id,
-    api_token: config.api_token,
+    api_token: loadEnvVar("EMBRACE_ORG_TOKEN", configPath),
```
jpmunz, line 115: maybe a console.error in the case this is blank with a message to set this in `.env`

.gitignore
```diff
@@ -96,4 +96,6 @@ RNTester/build
 nx.json
 
 # Yarn
-.yarn
\ No newline at end of file
+.yarn
+
+.env
```
jpmunz, line 101: I would have expected the naming to be flipped here, e.g. `.env` being what we commit as a default to the repo and `.env.local` being what gets gitignored which would be similar to how we have `settings/config.js` + `settings/config.local.js` setup in the dashjs repo

jpmunz (approved):
> @jpmunz I think the same can be done without using an extra package as dependency but just for quickness I introduced the dotenvx library to handle local variables using different env files. please, let me know if you have concerns about it

a couple comments but overall approach lgtm, I think fine to have an extra dependencies since it's just for development

package.json
```diff
@@ -28,6 +28,7 @@
     "@babel/preset-env": "^7.9.5",
     "@babel/preset-react": "^7.24.7",
     "@babel/preset-typescript": "^7.24.7",
+    "@dotenvx/dotenvx": "^1.48.3",
```
facostaembrace, line 31: @jpmunz moved the package into dev dependencies

package.json
facostaembrace: totally fair. I've moved it as dev dependency

jpmunz (approved):

PR #468, opened 2025-04-28
chore: fix dependabot warning

jpmunz:
Addressing a handful of dependabot warnings we had coming from the test-project folders: https://github.com/embrace-io/embrace-react-native-sdk/security/dependabot/114


jpmunz:
looks like dependabot figured out how to do this on its own: https://github.com/embrace-io/embrace-react-native-sdk/pull/469

PR #471, opened 2025-04-28
Placeholder Ruby code to un-glitch CodeQL

vitaliyf:


_attic/README.md
jpmunz, line 1: looks good, just adding this so we have the context later

jpmunz (approved):

Issue #473, opened 2025-04-30
[iOS] Upgrading from 6.0.0 to 6.0.1 breaks certain downloads


peterfajner-ensemble:
Hi, we have a React Native app that uses Embrace for logging/crash reporting as well as [Penthera](https://github.com/penthera/Download2Go-ios) for downloading media and [react-native-fs](https://github.com/itinance/react-native-fs) for downloading album art. After upgrading `@embrace-io/react-native` from `6.0.0` to `6.0.1`:

- On iOS, all Penthera flat-file (mp3) downloads result in a file of zero size. Charles logs show that the files were transferred fully.
- Also on iOS, all of `react-native-fs`'s network calls all hang forever.

It appears that some change in how Embrace hooks into the network stack breaks these downloads.

1. Is it possible to disable all network monitoring-related features of Embrace? We don't really use them and they've caused us issues in the past. We really only need crash analytics and manually-triggered logging.
2. Given that this update bumps the iOS Embrace version from 6.7.1 to 6.8.1, this should have been a minor version change, not a patch.

jpmunz:
@embrace-io/opensource-react-native for future reference, code snippet to reproduce the react-native-fs issue:

```typescript
import {readFile, downloadFile, DocumentDirectoryPath} from "react-native-fs";

function Reproduce(): React.JSX.Element {
  const onReactNativeFS = useCallback(async () => {
    const path = `${DocumentDirectoryPath}/some-file.txt`;
    console.log("PATH: ", path);

    try {
      const downloadResult = await downloadFile({
        fromUrl: 'https://github.githubassets.com/assets/primer-react.aef3ff9d485642192f52.module.css',
        toFile: path,
      });

      const finish = await downloadResult.promise;
      console.log("download result: ", finish);
    } catch (e) {
      console.log("download failed: ", e);
    }

    try {
      const fileResult = await readFile(path)
      console.log("file contents: ", fileResult);
    } catch (e) {
      console.log("file read failed: ", e);
    }

  }, []);

  return (
    <View>
      <Button title={"TEST react-native-fs"} onPress={onReactNativeFS} />
    </View>
  );
}
```

peterfajner-ensemble:
Thank you for the quick and detailed reply! Once the new RN SDK with the updated iOS dependency is available, I'll test with Penthera - I expect it's probably the same issue as with react-native-fs.

jpmunz:
[...]

PR #474, opened 2025-05-01
update iOS SDK dependency

jpmunz:
## Goal

Take advantage of fixes we are missing from the latest iOS SDK, specifically for network capture (see https://github.com/embrace-io/embrace-react-native-sdk/issues/473)


jpmunz:
looking into the GHA failure on the iOS tests, fails with just "Process completed with exit code 1." without much more details, the tests pass locally so far

PR #526, opened 2025-07-09
upgrade native sdk dependencies

jpmunz:
Upgrading Android dependency => 7.6.0


jpmunz:
will do the upgrade to iOS 6.12.1 separately since that involves some Podfile changes that require additional testing

PR #534, opened 2025-07-21
Add desugaring to app templates for lower Android API levels

jpmunz:
## Goal

Desugaring is required to support API levels 24 + 25 on Android. Setting this option necessitates a few additional changes on RN 0.71 + 0.72 since a higher version of AGP than they have by default


integration-tests/browserstack.conf.ts
```diff
@@ -14,8 +14,8 @@ const androidCapabilities = [
   },
   {
     "bstack:options": {
-      deviceName: "Samsung Galaxy S8",
```
jpmunz, line 17: BrowserStack dropped support for this device

facostaembrace (approved):

PR #709, opened 2026-02-10
[EMBR-4243] Add TurboModule (New Architecture) support to all native packages

alyssa-embrace:
Add React Native New Architecture (TurboModules) support to core, react-native-otlp, and react-native-tracer-provider while maintaining backward compatibility with the Old Architecture bridge modules.

- Add codegen spec files (NativeXxx.ts) for each package
- Add codegenConfig to each package.json
- Use TurboModuleRegistry with NativeModules fallback on JS side
- Add oldarch/newarch Android source sets with spec adapter classes
- Convert Android packages from ReactPackage to TurboReactPackage
- Rename iOS bridge files .m → .mm and add getTurboModule: categories
- Add Swift protocol conformance behind RCT_NEW_ARCH_ENABLED flag

## Notion ticket

https://www.notion.so/embraceio/Rewrite-packages-using-New-Architecture-c86220741f4244b2b79b36f38d349d0b

## Goal

Rewrite packages to integrate and fully utilize New Architecture

## Testing

Local testing and CI testing

## Release Notes

Full support for New Architecture with fallback

### DIFF
--- a/packages/core/src/EmbraceManagerModule.ts
+++ b/packages/core/src/EmbraceManagerModule.ts
@@ -6,8 +6,15 @@ const LINKING_ERROR =
   "- You rebuilt the app after installing the package\n" +
   "- You are not using Expo Go\n";
 
-export const EmbraceManagerModule = NativeModules.EmbraceManager
-  ? NativeModules.EmbraceManager
+// eslint-disable-next-line @typescript-eslint/no-explicit-any -- standard React Native TurboModule detection
+const isTurboModuleEnabled = (global as any).__turboModuleProxy != null;
+
+const NativeModule = isTurboModuleEnabled
+  ? require("./NativeEmbraceManager").default
+  : NativeModules.EmbraceManager;
+
+export const EmbraceManagerModule = NativeModule
+  ? NativeModule
   : new Proxy(
       {},
       {
--- a/packages/react-native-tracer-provider/src/NativeReactNativeTracerProviderModule.ts
+++ b/packages/react-native-tracer-provider/src/NativeReactNativeTracerProviderModule.ts
@@ -1,0 +1,33 @@
+import type {TurboModule} from "react-native";
+import {TurboModuleRegistry} from "react-native";
+
+/* eslint-disable @typescript-eslint/ban-types -- Object is required by React Native TurboModule codegen */
+export interface Spec extends TurboModule {
+  setupTracer(name: string, version: string, schemaUrl: string): void;
+  startSpan(
+    tracerName: string,
+    tracerVersion: string,
+    tracerSchemaUrl: string,
+    spanBridgeId: string,
+    name: string,
+    kind: string,
+    time: number,
+    attributes: Object,
+    links: Object[],
+    parentId: string,
+  ): Promise<Object>;
+  setAttributes(spanBridgeId: string, attributes: Object): void;
+  addEvent(
+    spanBridgeId: string,
+    eventName: string,
+    attributes: Object,
+    time: number,
+  ): void;
+  addLinks(spanBridgeId: string, links: Object[]): void;
+  setStatus(spanBridgeId: string, status: Object): void;
+  updateName(spanBridgeId: string, name: string): void;
+  endSpan(spanBridgeId: string, time: number): void;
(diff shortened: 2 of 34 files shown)


packages/core/src/EmbraceManagerModule.ts
```diff
@@ -6,8 +6,15 @@ const LINKING_ERROR =
   "- You rebuilt the app after installing the package\n" +
   "- You are not using Expo Go\n";
 
-export const EmbraceManagerModule = NativeModules.EmbraceManager
-  ? NativeModules.EmbraceManager
+// eslint-disable-next-line @typescript-eslint/no-explicit-any -- standard React Native TurboModule detection
```
jpmunz: is there an overall reference for the steps needed to make a turbo module backwards compatible that we could link to? I had found https://github.com/reactwg/react-native-new-architecture/blob/main/docs/backwards-compat-turbo-modules.md#unify-the-javascript-specs, its section on the typescript changes needed deviates from this but it maybe out-of-date as it hasn't been updated in a year

packages/react-native-tracer-provider/src/NativeReactNativeTracerProviderModule.ts
```diff
@@ -0,0 +1,33 @@
+import type {TurboModule} from "react-native";
+import {TurboModuleRegistry} from "react-native";
+
+/* eslint-disable @typescript-eslint/ban-types -- Object is required by React Native TurboModule codegen */
```
jpmunz: would this still complain if using the OTel types? E.g. `attributes: Attributes` ?

packages/core/src/EmbraceManagerModule.ts
alyssa-embrace: There's this, but it's deprecated: https://github.com/reactwg/react-native-new-architecture#guides

And this site links to the link you posted above: https://reactnative.dev/docs/turbo-native-modules-introduction

The final link ends up being this: https://github.com/reactwg/react-native-new-architecture/blob/main/docs/backwards-compat-turbo-modules.md

So I guess it isn't really out of date, just hasn't been updated in a hot minute.

alyssa-embrace: AFAIK the PR is backwards compatible at this time based on the listed docs.

alyssa-embrace: Is there a code change goal here?

packages/react-native-tracer-provider/src/NativeReactNativeTracerProviderModule.ts
alyssa-embrace: Based on this table it doesn't look like it: https://reactnative.dev/docs/appendix

`Attributes` is `AttributeValue | undefined` so it'd fall under the Union types listing.
There doesn't appear to be a native translation for that supported by turbomodules at this time.

force-ci
jpmunz, line 1: something like `git commit --allow-empty -m "force-ci"` should force the ci jobs to run without requiring a file change

packages/core/src/EmbraceManagerModule.ts
jpmunz: not the code itself but having those references available in comments so that we could confirm the approach against the recommendations

packages/react-native-tracer-provider/src/NativeReactNativeTracerProviderModule.ts
jpmunz: hmm I'm not sure I follow that reference, the note at the bottom says it strongly recommends object literals over using `Object` here, though there isn't a mapping shown in the table for the native side when using object literals in typescript

alyssa-embrace: Right. That was my bad; I missed that.

alyssa-embrace: I've altered the spec so that it should take object literals.

jpmunz:
since this has been open for a while going to close and use it as a baseline example to take a fresh pass

PR #809, opened 2026-05-06
[EMBR-11700] Expo Config Plugin: "Upload Debug Symbols" build phase references non-existent ${PODS_ROOT}/EmbraceIO/run.sh

### LINKED ISSUE #776: Expo Config Plugin: "Upload Debug Symbols" build phase references non-existent ${PODS_ROOT}/EmbraceIO/run.sh
## Description

The Expo Config Plugin's `withIosEmbraceAddUploadPhase` generates an Xcode build phase ("Upload Debug Symbols to Embrace") that hardcodes the path `${PODS_ROOT}/EmbraceIO/run.sh`. This file no longer exists in the EmbraceIO CocoaPod since the Apple SDK removed it in [6.10.0](https://github.com/embrace-io/embrace-apple-sdk/pull/240). The iOS build fails with:

```
/Users/.../ios/Pods/EmbraceIO/run.sh: No such file or directory
Command PhaseScriptExecution failed with a nonzero exit code
```

The `run.sh` script is bundled in the npm package at `node_modules/@embrace-io/react-native/ios/scripts/run.sh`, but the Expo plugin still points to the old pod path.

## Steps to Reproduce

1. Create an Expo project with `@embrace-io/react-native` 6.6.0
2. Configure the Embrace Expo plugin in `app.config.ts`:
3. Run `npx expo prebuild --clean`
4. Build for iOS

## Expected Behavior

The build phase should reference the correct path to `run.sh` — either `$SRCROOT/../node_modules/@embrace-io/react-native/ios/scripts/run.sh` or another valid location.

## Actual Behavior

Build fails because `${PODS_ROOT}/EmbraceIO/run.sh` does not exist.

## Environment

- `@embrace-io/react-native`: 6.6.0
- EmbraceIO pod resolved version: 6.16.2 (no longer ships `run.sh`)
- Expo SDK: 55
- React Native: 0.83.4
- Device: iPhone 16e, iOS 26.3
- Xcode: latest

alyssa-embrace:
## Goal

Fixes #776 - the `run.sh` script location has changed but the expo config plugin was still adding a build phase script using the old location.

Also fixes the fallback `createSourceMap` function in the `run.sh` script itself which also assumed the old script location - it now attempts to traverse up from the script directory until it finds the RN CLI and only tries to invoke the CLI if it finds it.

## Testing

Updated unit tests + tested manually with an Expo project

### DIFF
--- a/packages/core/ios/scripts/run.sh
+++ b/packages/core/ios/scripts/run.sh
@@ -105,7 +105,7 @@ function log_warning () {
 function run() {
     if [ -z "$EMBRACE_DEBUG_DSYM" ] ; then
         # we must redirect both stdout and stderr to /dev/null to have Xcode run this script in the background and move on to other build tasks
-	eval "$1 --log-path \"$BUILD_ROOT\"" > /dev/null 2>&1 &
+	eval "$1 --log-path $BUILD_ROOT" > /dev/null 2>&1 &
     else
         eval "$1"
     fi
@@ -190,7 +190,7 @@ function react_native_upload() {
     if [ -n "$react_args" ] ; then
         log "uploading react native resources with bundle at ${react_bundle_path} and map at ${react_map_path}"
         # shellcheck disable=2153
-        run "\"$UPLOAD_BIN_PATH\" --app \"$EMBRACE_ID\" --token \"$EMBRACE_TOKEN\" $react_args $host_arg --log-level $EMBRACE_LOG_LEVEL"
+        run "\"$UPLOAD_BIN_PATH\" --app $EMBRACE_ID --token $EMBRACE_TOKEN $react_args $host_arg --log-level $EMBRACE_LOG_LEVEL"
     fi
 }
 
@@ -270,7 +270,7 @@ function upload() {
         else
           icon_arg=""
         fi
-        run "\"$UPLOAD_BIN_PATH\" --app \"$EMBRACE_ID\" --token \"$EMBRACE_TOKEN\" $unity_args $icon_arg $app_version_arg $host_arg --dsym \"$app_dsym_file\" --log-level $EMBRACE_LOG_LEVEL"
+        run "\"$UPLOAD_BIN_PATH\" --app $EMBRACE_ID --token $EMBRACE_TOKEN $unity_args $icon_arg $app_version_arg $host_arg --dsym \"$app_dsym_file\" --log-level $EMBRACE_LOG_LEVEL"
     else
         log_warning "No app dSYM file was found under BUILD_ROOT ${BUILD_ROOT}. Skipping upload."
     fi
@@ -321,7 +321,7 @@ function upload() {
             log "Queueing framework dSYM for upload: $dsym_full_path"
         done
         IFS=$ORIG_IFS
-        run "\"$UPLOAD_BIN_PATH\" --app \"$EMBRACE_ID\" --token \"$EMBRACE_TOKEN\" $dsym_args $app_version_arg $host_arg --log-level $EMBRACE_LOG_LEVEL"
+        run "\"$UPLOAD_BIN_PATH\" --app $EMBRACE_ID --token $EMBRACE_TOKEN $dsym_args $app_version_arg $host_arg --log-level $EMBRACE_LOG_LEVEL"
     fi
 }
 
--- a/packages/core/ios/scripts/run.sh
+++ b/packages/core/ios/scripts/run.sh
@@ -105,14 +105,18 @@ function log_warning () {
 function run() {
     if [ -z "$EMBRACE_DEBUG_DSYM" ] ; then
         # we must redirect both stdout and stderr to /dev/null to have Xcode run this script in the background and move on to other build tasks
-	eval "$1 --log-path $BUILD_ROOT" > /dev/null 2>&1 &
+	eval "$1 --log-path \"$BUILD_ROOT\"" > /dev/null 2>&1 &
     else
         eval "$1"
     fi
 }
 
 function createSourceMap() {
-  REACT_NATIVE_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/../../.." && pwd)"
+  REACT_NATIVE_DIR="$(dirname "${BASH_SOURCE[0]}")"
+  while [ "$REACT_NATIVE_DIR" != "/" ]; do
+    [ -f "$REACT_NATIVE_DIR/node_modules/react-native/cli.js" ] && break
+    REACT_NATIVE_DIR="$(dirname "$REACT_NATIVE_DIR")"
+  done
   CLI_PATH="$REACT_NATIVE_DIR/node_modules/react-native/cli.js"
 
   DEST=/dev/null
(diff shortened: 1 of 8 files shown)


packages/core/ios/scripts/run.sh
jpmunz, line 1: we should avoid drift since this script was copied over from the internal go repo and instead make changes there first and then copy over once they're merged

also could you enumerate what the particular changes in this file are fixing?

facostaembrace: +1

packages/core/src/plugin/withIosEmbrace.ts
```diff
@@ -296,15 +296,18 @@ const withIosEmbraceAddUploadPhase: ConfigPlugin<EmbraceProps> = (
     /*
     shellScript = "REACT_NATIVE_MAP_PATH=\"$CONFIGURATION_BUILD_DIR/embrace-assets/main.jsbundle.map\" EMBRACE_ID=ios789 EMBRACE_TOKEN=apiToken456 \"${PODS_ROOT}/EmbraceIO/run.sh\"\nrm \"$CONFIGURATION_BUILD_DIR/embrace-assets/main.jsbundle.map\"";
      */
-    if (!findPhase(project, "EmbraceIO/run.sh")) {
+    if (
+      !findPhase(project, "EmbraceIO/run.sh") &&
```
facostaembrace: I would like to adjust and improve this by moving these two paths outside, into a constant. In this way it won't be missed across the the project and it would be reusable in all places

packages/core/src/plugin/withIosEmbrace.ts
```diff
@@ -296,15 +296,18 @@ const withIosEmbraceAddUploadPhase: ConfigPlugin<EmbraceProps> = (
...
+    ) {
       project.addBuildPhase(
         [],
         "PBXShellScriptBuildPhase",
         "Upload Debug Symbols to Embrace",
         null,
         {
           shellPath: "/bin/sh",
-          shellScript: `REACT_NATIVE_MAP_PATH="${sourceMapPath}" EMBRACE_ID=${props.iOSAppId} EMBRACE_TOKEN=${props.apiToken} "\${PODS_ROOT}/EmbraceIO/run.sh"`,
+          shellScript: `REACT_NATIVE_MAP_PATH="${sourceMapPath}" EMBRACE_ID=${props.iOSAppId} EMBRACE_TOKEN=${props.apiToken} "\${SRCROOT}/../node_modules/@embrace-io/react-native/ios/scripts/run.sh"`,
```
facostaembrace: also here I would like to find a way to avoid hardcoding this location. maybe the right location (the `../node_modules/` part) can be calculated at the top of the file and then be passed here. this would prevent the script to be broken again in case the path changes its location

packages/core/src/plugin/withIosEmbrace.ts
```diff
@@ -296,15 +296,18 @@ const withIosEmbraceAddUploadPhase: ConfigPlugin<EmbraceProps> = (
     /*
     shellScript = "REACT_NATIVE_MAP_PATH=\"$CONFIGURATION_BUILD_DIR/embrace-assets/main.jsbundle.map\" EMBRACE_ID=ios789 EMBRACE_TOKEN=apiToken456 \"${PODS_ROOT}/EmbraceIO/run.sh\"\nrm \"$CONFIGURATION_BUILD_DIR/embrace-assets/main.jsbundle.map\"";
      */
-    if (!findPhase(project, "EmbraceIO/run.sh")) {
+    if (
+      !findPhase(project, "EmbraceIO/run.sh") &&
+      !findPhase(project, "embrace-io/react-native/ios/scripts/run.sh")
```
facostaembrace: question: can we confirm we don't need the `@` here? (`@embrace-io/react-native/ios/scripts/run.sh`)

facostaembrace:
@alyssa-embrace JS tests still fail, I think they need attention

packages/core/src/plugin/textUtils.ts
```diff
@@ -1,5 +1,5 @@
-const hasMatch = (lines: string[], matcher: string) => {
-  return lines.find(line => line.match(matcher));
+const hasMatch = (lines: string[], substring: string) => {
+  return lines.find(line => line.includes(substring));
```
hamudi-ahmed, line 2: `String.match()` coerces string args into a regex - so any regex special characters in the passed in string (e.g. the leading $ in the new run-script path) were being misinterpreted. Switched to `String.includes()` since this is what all callers are actually trying to do anyway.

packages/core/src/plugin/withIosEmbrace.ts
```diff
@@ -294,17 +296,20 @@ const withIosEmbraceAddUploadPhase: ConfigPlugin<EmbraceProps> = (
     }
 
     /*
-    shellScript = "REACT_NATIVE_MAP_PATH=\"$CONFIGURATION_BUILD_DIR/embrace-assets/main.jsbundle.map\" EMBRACE_ID=ios789 EMBRACE_TOKEN=apiToken456 \"${PODS_ROOT}/EmbraceIO/run.sh\"\nrm \"$CONFIGURATION_BUILD_DIR/embrace-assets/main.jsbundle.map\"";
+    shellScript = "REACT_NATIVE_MAP_PATH=\"$CONFIGURATION_BUILD_DIR/embrace-assets/main.jsbundle.map\" EMBRACE_ID=ios789 EMBRACE_TOKEN=apiToken456 \"${SRCROOT}/../node_modules/@embrace-io/react-native/ios/scripts/run.sh\"\nrm \"$CONFIGURATION_BUILD_DIR/embrace-assets/main.jsbundle.map\"";
      */
-    if (!findPhase(project, "EmbraceIO/run.sh")) {
+    if (
+      !findPhase(project, embraceLegacyRunScriptPath) &&
+      !findPhase(project, embraceRunScriptPath)
```
hamudi-ahmed, line 304: note: this check prevents us from adding an additional upload phase if the old one already exists, but it doesn't 'fix' an existing xcode project by replacing the build phase - customers with existing/broken build phases will need to fix them manually or run prebuild with the `--clean` flag

packages/core/src/plugin/withIosEmbrace.ts
hamudi-ahmed: Updated

packages/core/src/plugin/withIosEmbrace.ts
hamudi-ahmed: Updated

packages/core/src/plugin/withIosEmbrace.ts
hamudi-ahmed: Technically it's not needed here, but this has been moved into a constant using the full path anyway

packages/core/ios/scripts/run.sh
hamudi-ahmed: Updated the PR description and raised a PR on the go repo to mirror the changes made here

jpmunz (approved):

PR #812, opened 2026-05-11
[EMBR-4221] Publish React Native SDK via CI/CD

alyssa-embrace:
## Summary

  Replaces the manual local publish process with a fully automated CI/CD pipeline triggered on every merge to main.

  - Post-merge pipeline (post-merge-main.yml): runs all unit tests (JS, iOS, Android) and integration tests sequentially on main, then runs a dry-run npm publish. If everything
  passes, automatically creates a draft GitHub Release with the proposed next MINOR version and commit notes since the last release.
  - Release workflow (release-npm.yml): a workflow_dispatch button that publishes to npm with --provenance, pushes the version tag to main, and publishes (or creates) the GitHub
  Release. The version can be overridden at trigger time — the draft release body links directly to this workflow with the proposed version pre-suggested.
  - Reusable workflows: added workflow_call support to node.js.yml, android.unit-test.yml, ios.unit-test.yml, and integration-test.yml so the post-merge orchestrator can call them
   and chain them with needs. A force: boolean input on the iOS and Android workflows bypasses the paths-filter skip logic so all tests always run on main, regardless of which
  files changed in a given push.

  Release flow

  merge to main
    → unit tests (JS + iOS + Android) in parallel
    → integration tests (full matrix, 10 apps × 2 platforms)
    → npm dry run
    → draft GitHub Release created with proposed vX.(Y+1).0 + commit notes

  developer reviews draft → clicks "Run workflow" on release-npm.yml
    → version bump committed + tagged → pushed to main
    → npm publish --provenance
    → draft GitHub Release published

  Test plan

  - Merge a test commit to main and confirm the post-merge pipeline triggers and runs all stages in order
  - Confirm a draft GitHub Release is created with the correct proposed MINOR version and recent commit notes
  - Confirm the pipeline fails fast at the appropriate stage if any test or dry-run fails (no draft release created)
  - Trigger release-npm.yml via workflow_dispatch with a test version and confirm the tag is pushed, npm publish runs, and the GH release is updated
  - Confirm existing PR workflows (node.js.yml, android.unit-test.yml, ios.unit-test.yml) are unaffected — paths-filter skip logic still works as before on PRs

  Notes

  - release-npm.yml does a direct push to main for the version bump commit and tag. If branch protection requires PRs, add a GitHub Actions bypass rule in branch protection
  settings.
  - Integration tests still gate on release/ and embrace-ci/ branch prefixes for PR triggers — only the workflow_call and workflow_dispatch paths skip that check.


.github/workflows/publish.yml
```diff
@@ -0,0 +1,102 @@
+name: Publish to npm
```
facostaembrace: This file is never called by any other workflow in this PR — `post-merge-main.yml` implements the dry-run inline, and `release-npm.yml` does the actual publish directly. As written it's unreachable dead code. Was the intent for `post-merge-main.yml` to delegate its dry-run here, or for `release-npm.yml` to call this? Either wire it in or remove it before merging.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,102 @@
...
+permissions:
+  contents: write
+  id-token: write
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
```
facostaembrace: These use floating version tags (`@v4`) while the rest of the repo pins to commit SHAs — see `release-npm.yml` and `post-merge-main.yml` in this very PR. The repo already runs zizmor to enforce pinned SHAs. Please pin both `actions/checkout` and `actions/setup-node` to their SHA equivalents (same ones used in the other new workflows).

.github/workflows/release-npm.yml
```diff
@@ -0,0 +1,83 @@
...
+  id-token: write
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          ref: main
```
facostaembrace: This hardcodes `main` as the checkout ref. DEVELOPING.md explicitly calls out hotfix releases: _"urgent patch fixes: branch from the latest released tag, cut a patch release from that new branch."_ With this hardcoded, a hotfix on another branch can't be published without editing the workflow. Could you add a `ref` input (default: `main`) to `workflow_dispatch`, the same way `publish.yml` does it?

.github/workflows/release-npm.yml
```diff
@@ -0,0 +1,83 @@
...
+          fetch-depth: 0
+
+      - uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6.3.0
+        with:
+          node-version-file: ".node-version"
+          registry-url: "https://registry.npmjs.org"
+
+      - run: corepack enable
+
+      - run: yarn
```
facostaembrace: Suggestion: add an `npm whoami` step right after `yarn install` to validate that `NPM_TOKEN` is authenticated and scoped for `@embrace-io` before running the full build and test suite. If the token is expired or misconfigured, right now you'd only find out after ~10 minutes of CI time.

```yaml
- name: Verify npm credentials
  run: npm whoami
  env:
    NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```

.github/workflows/release-npm.yml
```diff
@@ -0,0 +1,83 @@
...
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+
+      - name: Bump version and create git tag
+        run: npx lerna version ${{ inputs.version }} --yes --no-push
+
+      - run: yarn build
+
+      - run: yarn test --watchAll=false
+
+      - name: Push version commit and tags to main
```
facostaembrace: The comment here acknowledges the branch protection requirement, but this needs to be a hard prerequisite, not a reminder. The first real release attempt will fail at this step if it isn't already set up. Can we confirm this is resolved _before_ this PR merges, not after?

.github/workflows/release-npm.yml
```diff
@@ -0,0 +1,83 @@
...
+        # Requires 'contents: write' and main branch protection configured to allow
+        # GitHub Actions pushes (or no branch protection requiring PRs).
+        run: git push --follow-tags
+
+      - name: Publish to npm
+        run: npx lerna publish from-git --yes --dist-tag ${{ inputs.dist-tag }} --provenance
```
facostaembrace: DEVELOPING.md step 8 uses `update-embrace-package.sh` to point a test app at the freshly published npm artifact. The integration tests that run in `post-merge-main.yml` exercise the workspace source — not the published tarball. A packaging mistake (wrong `main` field, missing files in the tarball) would ship undetected. Worth calling this out as a known gap, even if it's addressed as a follow-up.

.github/workflows/post-merge-main.yml
```diff
@@ -0,0 +1,118 @@
...
+
+      - run: yarn
+
+      - name: Compute proposed MINOR version
+        id: version
+        run: |
+          CURRENT=$(node -p "require('./lerna.json').version")
+          MAJOR=$(echo "$CURRENT" | cut -d. -f1)
+          MINOR=$(echo "$CURRENT" | cut -d. -f2)
+          echo "proposed=${MAJOR}.$((MINOR + 1)).0" >> "$GITHUB_OUTPUT"
```
facostaembrace: The proposed version is always `MAJOR.(MINOR+1).0`. Past releases include patch versions (`v6.2.1`, `v6.4.1`), and DEVELOPING.md has a documented hotfix path. When a patch is needed, this draft will consistently show the wrong proposed version — confusing, even if the actual release step lets you override it.

Two options:
- **a)** Add a `workflow_dispatch` trigger to this workflow with a `bump` input (`patch | minor | major`, default `minor`) so the draft can be re-triggered with the right type.
- **b)** At minimum, make the override instruction in the draft body more prominent (e.g. bold, first bullet).

.github/workflows/post-merge-main.yml
```diff
@@ -0,0 +1,118 @@
...
+
+      - name: Set version (local only)
+        run: npx lerna version ${{ steps.version.outputs.proposed }} --yes --no-push
+
+      - run: yarn build
```
facostaembrace: The `js-unit-tests` job (which calls `node.js.yml`) already runs `yarn build` and `yarn test`. Running them again here adds ~5 minutes to every main merge. The unique value of this dry-run step is the version bump + `npm pack --dry-run` — could the duplicate `yarn build` and `yarn test` lines be removed, leaving only the pack step after bumping?

.github/workflows/post-merge-main.yml
```diff
@@ -0,0 +1,118 @@
...
+
+$(cat /tmp/commit-list.txt)
+
+---
+
+**To publish this release:**
+1. Go to [Actions → Release to npm](${WORKFLOW_URL})
+2. Click **Run workflow**
+3. Enter version \`${PROPOSED_VERSION}\` (or override with your preferred version)
+4. Click **Run workflow** — this publishes to npm with provenance and creates the GitHub release
```
facostaembrace: The publish instructions are missing step 9 from DEVELOPING.md: updating the external changelog at `embrace-io/embrace-docs`. This is easy to forget after a successful publish. Even a simple line at the end — e.g. `5. Update the [external changelog](https://github.com/embrace-io/embrace-docs/blob/main/docs/react-native/changelog.md)` — would prevent it being silently skipped.

jpmunz:
Question about the overall workflow, so we have:

1. merge to master -> draft GitHub Release created with proposed vX.(Y+1).0 + commit notes
2. developer reviews draft → clicks "Run workflow" on release-npm.yml
3. → version bump committed + tagged → pushed to main
4. → npm publish --provenance
5. → draft GitHub Release published

After #1 then we'll have a draft GH release but the version bump isn't happening until #3, would that mean if someone published the draft release manually before #3 that we would do a publish to npm with the updated commits since last release but the old version numbers?

.github/workflows/publish.yml
alyssa-embrace: Fixed, thank you.

.github/workflows/release-npm.yml
alyssa-embrace: Done

.github/workflows/release-npm.yml
alyssa-embrace: Added!

.github/workflows/release-npm.yml
alyssa-embrace: Definitely. I'll loop in @vitaliyf

alyssa-embrace:
I believe so @jpmunz

jpmunz:
> I believe so @jpmunz

hmm how does that compare to the flow on the android + apple repos? It seems a bit odd to have a draft release that wouldn't be valid to publish until a separate action is triggered

.github/workflows/release-npm.yml
alyssa-embrace: Noted

alyssa-embrace:
@jpmunz I talked with Ahmed about it, and we agreed. So we moved to a `npm pack` and artefact upload as our so-called `staging` release. With the new flow, we download the tarball from the action and then manually map that by serving it from a local NPM service for testing. Alternatively we can point to it as an artefact for automated testing as well.

hamudi-ahmed:
After taking a closer look and reviewing some of the other repos' workflows, I have a few concerns with the current approach:

- everything happens directly on main  - this means it will need to bypass branch protection rules, and there's no opportunity for a human to review the actual version bump commit before it's pushed to main and published. 
- currently the two workflows are completely decoupled - there's no guard against someone inadvertently triggering the release workflow if the post-merge-main workflow fails (unlikely in practice, but possible).
- the git tag is pushed before the publish happens - so if the publish step fails for any reason, there's a dangling tag that needs cleaning up.
- The packing and uploading of artifacts in `post-merge-main.yml` doesn't add a lot of value IMO - since the version changes are done separately in each workflow there's no guarantee that what you're testing is actually what will be published.

I think it would be better to adopt the general approach we have in other repos where the initial workflow creates a release branch and PR into main with the version bump changes, and then a publish workflow that is triggered when the release PR is approved and merged:

Stage 1 - prepare release
- Triggered manually (`workflow_dispatch`) with inputs for the proposed version bump
- Checks out `main` and runs lerna to bump versions without pushing or creating a tag (`lerna version --no-git-tag-version --no-push`)
- Pushes the changes to a release branch (e.g. `release/vx.x.x`) and creates a release PR into `main`
- The full test suite then runs against this release branch/PR. Any manual testing can also be done on the release PR at this stage if required.

Stage 2 - publish
- Triggered only when a release PR is merged (we could gate this on a PR label or commit message, for example)
- Checks out the merged commit and runs a build
-  Publishes whatever the current version using `lerna publish from-package` (the version bump happens in Stage 1)
- Creates and pushes the tag (only happens if the publish succeeds)
- Creates a GitHub release pointing at the new tag.

alyssa-embrace:
Closed as I think that @hamudi-ahmed's branch is better.

PR #823, opened 2026-05-15
[EMBR-12207] remove expo-router dependencies from non-expo test templates

hamudi-ahmed:
## Goal

Stop pulling `expo-router` into the non-expo integration test templates. These templates were failing to build because of conflicting transitive dependencies pulled in by the latest Expo version (via `expo-router`). None of them actually use `expo-router` - it was only listed as a dependency for build-time module resolution because the test harness supports both navigation libraries (expo-router and react-navigation)

This PR splits the test harness into separate entrypoints for React Native and Expo, and makes `expo-router` an optional peer dependency there, allowing us to drop `expo-router` from those templates entirely

Also updates the github workflow to allow the tests to be triggered manually

## Testing

Manually triggered the GH action to verify that the tests now pass


integration-tests/test-harness/src/index.ts
```diff
@@ -1,4 +1,4 @@
-export * from "./EmbraceTestHarness";
+export * from "./EmbraceReactNativeTestHarness";
```
jpmunz, line 1: could `EmbraceExpoTestHarness` also be exported here while preserving the separate entrypoint? And then removing the extra `lib` portion from `import {EmbraceExpoTestHarness} from '@embrace-io/react-native-test-harness/lib/EmbraceExpoTestHarness';`?

jpmunz (approved):
just a small question about the export but overall lgtm!

integration-tests/test-harness/src/index.ts
hamudi-ahmed: unfortunately not - re-exporting `EmbraceExpoTestHarness` here means you end up with a static `require` that forces Metro to walk the whole dependency chain - it doesn't treeshake it out. I considered doing something with package.json exports, but Metro doesn't support that in RN 71, so decided to go with the simpler (albeit slightly ugly) approach.

PR #845, opened 2026-05-27
[EMBR-4221] Publish from CI

alyssa-embrace:
## Goal

Update release process to publish from CI

### DIFF
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -70,8 +70,8 @@ jobs:
 
       - name: Configure git identity
         run: |
-          git config user.name "github-actions[bot]"
-          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git config user.name "embrace-ci[bot]"
+          git config user.email "embrace-ci@users.noreply.github.com"
 
       - name: Resolve bump argument
         id: bump
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -56,9 +56,10 @@ jobs:
           fetch-depth: 1
 
       - name: Setup Node
-        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0 # zizmor: ignore[cache-poisoning]
         with:
           node-version-file: ".node-version"
+          cache: ''
 
       - name: Enable Corepack (Yarn 4)
         run: corepack enable
@@ -120,5 +121,10 @@ jobs:
               --body "Automated release PR opened by @$ACTOR (bump=\`$BUMP\`).
 
           Merging this PR triggers \`release.yml\`, which publishes \`@embrace-io/react-native@$VERSION\` and the four sibling packages to npm.
+
+          **Post-merge checklist:**
+          - [ ] Verify packages appear on https://www.npmjs.com/org/embrace-io
+          - [ ] Confirm the published package runs on the latest supported version of React Native
+          - [ ] Update the [external changelog](https://github.com/embrace-io/embrace-docs/blob/main/docs/react-native/changelog.md)
 
           To abort: close this PR without merging."
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -57,10 +57,10 @@ jobs:
           persist-credentials: false
 
       - name: Setup Node
-        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0 # zizmor: ignore[cache-poisoning]
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0 # zizmor: ignore[cache-poisoning] - job doesn't produce published artifacts
         with:
           node-version-file: ".node-version"
-          cache: 'npm'
+          cache: 'yarn'
 
       - name: Enable Corepack (Yarn 4)
         run: corepack enable
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -115,14 +115,14 @@ jobs:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           VERSION: ${{ steps.version.outputs.value }}
           BUMP: ${{ inputs.bump }}
-          ACTOR: ${{ github.actor }}
         run: |
           gh pr create \
               --base main \
               --head "release/v$VERSION" \
               --title "release: v$VERSION" \
               --label "release" \
-              --body "Automated release PR opened by @$ACTOR (bump=\`$BUMP\`).
+              --assignee "$GITHUB_ACTOR" \
+              --body "Automated release PR opened by @$GITHUB_ACTOR (bump=\`$BUMP\`).
 
           Merging this PR triggers \`release.yml\`, which publishes \`@embrace-io/react-native@$VERSION\` and the four sibling packages to npm.
 
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -54,12 +54,13 @@ jobs:
         with:
           ref: main
           fetch-depth: 1
+          persist-credentials: false
 
       - name: Setup Node
         uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0 # zizmor: ignore[cache-poisoning]
         with:
           node-version-file: ".node-version"
-          cache: ''
+          cache: 'npm'
 
       - name: Enable Corepack (Yarn 4)
         run: corepack enable
@@ -99,11 +100,14 @@ jobs:
 
       - name: Create release branch and push
         env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_REPOSITORY: ${{ github.repository }}
           VERSION: ${{ steps.version.outputs.value }}
         run: |
           git checkout -b "release/v$VERSION"
           git add -A
           git commit -m "release: v$VERSION"
+          git remote set-url origin "https://x-access-token:${GITHUB_TOKEN}@github.com/${GITHUB_REPOSITORY}.git"
           git push origin "release/v$VERSION"
 
       - name: Open release PR
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -1,0 +1,124 @@
+name: Prepare Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      bump:
+        description: 'Type of version bump'
+        type: choice
+        required: true
+        options:
+          - patch
+          - minor
+          - major
+          - custom
+      custom_version:
+        description: 'Required if bump == custom (e.g. 6.7.0-beta.1, 6.6.1)'
+        type: string
+        required: false
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  prepare:
+    name: Open release PR
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - name: Validate inputs
+        env:
+          BUMP: ${{ inputs.bump }}
+          CUSTOM: ${{ inputs.custom_version }}
+        run: |
+          if [ "$BUMP" = "custom" ]; then
+            if [ -z "$CUSTOM" ]; then
+              echo "::error::custom_version is required when bump=custom" >&2
+              exit 1
+            fi
+            # Basic semver shape check: X.Y.Z optionally followed by -prerelease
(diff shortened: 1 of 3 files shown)


.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,130 @@
...
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          ref: main
+          fetch-depth: 1
+
+      - name: Setup Node
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0 # zizmor: ignore[cache-poisoning]
+        with:
+          node-version-file: ".node-version"
+          cache: ''
```
jpmunz: why not use 'npm' as the cache here?

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,130 @@
...
+      - name: Enable Corepack (Yarn 4)
+        run: corepack enable
+
+      - name: Install dependencies
+        run: yarn install --immutable
+
+      - name: Configure git identity
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
jpmunz: curious about the magic number here, it seems like we have it in some spots but not others e.g.:
https://github.com/embrace-io/release-drafter/blob/693d20e7c1ce1a81d3a41962f85914253b518449/.github/workflows/licensed.yml#L61
https://github.com/embrace-io/release-drafter/blob/693d20e7c1ce1a81d3a41962f85914253b518449/.github/workflows/renovate-build.yml#L32

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,130 @@
...
+      - name: Open release PR
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          VERSION: ${{ steps.version.outputs.value }}
+          BUMP: ${{ inputs.bump }}
+          ACTOR: ${{ github.actor }}
+        run: |
+          gh pr create \
+              --base main \
+              --head "release/v$VERSION" \
```
jpmunz, line 121: does this also need `-a "$ACTOR"` ?

.github/workflows/prepare-release.yml
jpmunz: and actually here we use embrace-ci@users.noreply.github.com:
https://github.com/embrace-io/embrace-web-sdk/blob/main/.github/workflows/release.yaml#L65

should confirm with devops which they'd prefer

.github/workflows/release.yml
```diff
@@ -0,0 +1,102 @@
...
+    timeout-minutes: 30
+    steps:
+      - name: Checkout merge commit
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          ref: ${{ github.event.pull_request.merge_commit_sha }}
+          fetch-depth: 1
+
+      - name: Setup Node + npm auth
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0 # zizmor: ignore[cache-poisoning]
```
jpmunz: shouldn't ignore for this step in particular since it is doing the actual publish, can also add `package-manager-cache: false` to the `with` list

.github/workflows/release.yml
```diff
@@ -0,0 +1,102 @@
...
+          cache: ''
+
+      - name: Enable Corepack (Yarn 4)
+        run: corepack enable
+
+      - name: Install dependencies
+        run: yarn install --immutable
+
+      - name: Build all packages
+        run: npx lerna run build
```
jpmunz: either way works but maybe slight preference for `yarn build` since we have that defined as a script in package.json in case we ever decide to change the associated command

.github/workflows/release.yml
```diff
@@ -0,0 +1,102 @@
...
+
+      - name: Read version and determine dist-tag
+        id: version
+        run: |
+          VERSION=$(node -p "require('./lerna.json').version")
+          case "$VERSION" in
+            *-*) DIST_TAG=next ;;
```
jpmunz, line 50: maybe worth a quick comment explaining the matching here, this is catching a custom release candidate version vs. a stable one?

.github/workflows/release.yml
```diff
@@ -0,0 +1,102 @@
...
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+          DIST_TAG: ${{ steps.version.outputs.dist_tag }}
+        run: |
+          npx lerna publish from-package --yes --dist-tag "$DIST_TAG" --provenance
+
+
+      - name: Configure git identity
+        run: |
+          git config user.name "embrace-ci[bot]"
```
jpmunz, line 70: should probably be consistent with the identity used in prepare-release

.github/workflows/release.yml
```diff
@@ -0,0 +1,102 @@
...
+          echo "value=$VERSION" >> "$GITHUB_OUTPUT"
+          echo "dist_tag=$DIST_TAG" >> "$GITHUB_OUTPUT"
+          echo "Version: $VERSION  dist-tag: $DIST_TAG"
+
+      - name: Publish to npm
+        # from-package is idempotent for partial failures, but exits non-zero if all packages
+        # are already published. If this step fails after a prior successful publish, recover
+        # manually — see DEVELOPING.md § "Release workflow: publish succeeds but tag or GitHub release was not created"
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
jpmunz: is this needed? It seems like we don't need to supply it when publishing the web-sdk:
https://github.com/embrace-io/embrace-web-sdk/blob/main/.github/workflows/ci-nodejs.yml#L742

I'd check w/ devops if this is somehow automatically being supplied

.github/workflows/release.yml
```diff
@@ -0,0 +1,102 @@
...
+        run: |
+          npx lerna publish from-package --yes --dist-tag "$DIST_TAG" --provenance
+
+
+      - name: Configure git identity
+        run: |
+          git config user.name "embrace-ci[bot]"
+          git config user.email "embrace-ci[bot]@users.noreply.github.com"
+
+      - name: Push git tag
```
jpmunz, line 73: does lerna already do this as part of its `publish` command?

.github/workflows/release.yml
```diff
@@ -0,0 +1,102 @@
...
+          MERGE_SHA: ${{ github.event.pull_request.merge_commit_sha }}
+        run: |
+          if git ls-remote --tags origin "refs/tags/v$VERSION" | grep -q .; then
+            echo "Tag v$VERSION already exists, skipping."
+          else
+            git tag "v$VERSION" "$MERGE_SHA"
+            git push origin "v$VERSION"
+          fi
+
+      - name: Create GitHub Release
```
jpmunz, line 88: the other workflow option would be to have only a Draft Release created on merging the release/ branch and then the actual publish action triggered by hitting Publish on that Draft release, we can discuss separately

jpmunz (commented):
looks like `persist-credentials: false` is needed for a few of the steps to deal with the zizmor warnings

.github/workflows/release.yml
hamudi-ahmed: good catch - this is assuming that we're using tokens to publish. Looks like the web SDK switched to using OIDC some time ago - we should do the same here. I'll reach out to devops team.

One thing to note: OIDC requires [lerna v9](https://lerna.js.org/docs/recipes/oidc-trusted-publishing) and [npm v11](https://docs.npmjs.com/trusted-publishers) - so we'll need to bump lerna and either update the repo to node 24 or manually install npm v11 in the GHA (since node 22 bundles npm v10)

.github/workflows/release.yml
```diff
@@ -0,0 +1,102 @@
...
+
+      - name: Publish to npm
+        # from-package is idempotent for partial failures, but exits non-zero if all packages
+        # are already published. If this step fails after a prior successful publish, recover
+        # manually — see DEVELOPING.md § "Release workflow: publish succeeds but tag or GitHub release was not created"
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+          DIST_TAG: ${{ steps.version.outputs.dist_tag }}
+        run: |
+          npx lerna publish from-package --yes --dist-tag "$DIST_TAG" --provenance
```
hamudi-ahmed: Turns out `lerna publish` doesn't support the `--provenance` flag - I think we'll need to use one of these methods to publish with provenance (environment variable maybe?): https://docs.npmjs.com/generating-provenance-statements#using-third-party-package-publishing-tools

jpmunz: ya I like the `NPM_CONFIG_PROVENANCE` option to make it clear the provenance should only be applied when publishing from this workflow specifically

.github/workflows/prepare-release.yml
alyssa-embrace: The goal as I understand it is to disable caching; so this is meant to explicitly opt out.

.github/workflows/release.yml
alyssa-embrace: Noted, I'll make that edit.

.github/workflows/prepare-release.yml
jpmunz: if we want to disable caching for this step we should remove the ` # zizmor: ignore[cache-poisoning]` ignore, would also need `package-manager-cache: false` (the default for `cache` is '' already so that can be removed)

for this particular job though since there are no artifacts being published and this is just generating the PR for the version bump we should be fine keeping `zizmor: ignore[cache-poisoning]` and setting 'npm' as the cache since there wouldn't be anything to attack with a cache poisoning

alyssa-embrace:
Noted and added `persist-credentials: false`

.github/workflows/release.yml
alyssa-embrace: It doesn't from what I can tell. We suppress tag creation on `lerna version` though in `prepare-release.yml`

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,134 @@
...
+      - name: Create release branch and push
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_REPOSITORY: ${{ github.repository }}
+          VERSION: ${{ steps.version.outputs.value }}
+        run: |
+          git checkout -b "release/v$VERSION"
+          git add -A
+          git commit -m "release: v$VERSION"
+          git remote set-url origin "https://x-access-token:${GITHUB_TOKEN}@github.com/${GITHUB_REPOSITORY}.git"
```
jpmunz, line 110: is this needed? I don't see it being set in https://github.com/embrace-io/embrace-web-sdk/blob/main/.github/workflows/release.yaml#L68 for example

.github/workflows/release.yml
```diff
@@ -0,0 +1,108 @@
...
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+
+      - name: Push git tag
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_REPOSITORY: ${{ github.repository }}
+          VERSION: ${{ steps.version.outputs.value }}
+          MERGE_SHA: ${{ github.event.pull_request.merge_commit_sha }}
+        run: |
+          git remote set-url origin "https://x-access-token:${GITHUB_TOKEN}@github.com/${GITHUB_REPOSITORY}.git"
```
jpmunz, line 80: similar question about the set-url being needed here

.github/workflows/prepare-release.yml
alyssa-embrace: My understanding is that we do need it explicitly because we're using `persist-credentials: false`. I checked the other workflow, and it's using `persist-credentials: true`.

I think we'll fail auth if we don't use either or and the workflow will fail.

jpmunz: ah ok that makes sense

.github/workflows/release.yml
hamudi-ahmed: lerna and node have now been updated in main, and devops have set up OIDC for npm, so this should be good to go

hamudi-ahmed: (as in we can just remove the NODE_AUTH_TOKEN)

.github/workflows/prepare-release.yml
hamudi-ahmed: the magic number is correct for the github actions bot, but you're right, this is inconsistent with what we have in the web SDK, and even `update-native-sdks.yaml` in this repo, which both use `embrace-ci` for the git identity. I've updated this to match those, so the version bump commit and git tag will both be attributed to `embrace-ci`.

.github/workflows/prepare-release.yml
hamudi-ahmed: Updated

.github/workflows/release.yml
hamudi-ahmed: Removed the npm token, so this should all now be correctly set up to use OIDC

.github/workflows/release.yml
jpmunz: will consider as a follow-up

jpmunz (approved):

facostaembrace (approved):

PR #877, opened 2026-06-12
[EMBR-12601] Update unit test projects to react native 0.85

hamudi-ahmed:
## Goal

Updates the unit test projects to React Native 0.85 in order to clear various dependabot security alerts caused by transitive RN dependencies (note: these vulnerabilities are just in test project dependencies and don't affect any code that is actually shipped).

## Testing

Ran the native Android/iOS unit tests

### DIFF
--- a/packages/react-native-tracer-provider/test-project/ios/RNEmbraceTracerProviderTests/RNEmbraceTracerProviderTests.swift
+++ b/packages/react-native-tracer-provider/test-project/ios/RNEmbraceTracerProviderTests/RNEmbraceTracerProviderTests.swift
@@ -87,8 +87,6 @@ class ReactNativeTracerProviderTests: XCTestCase {
   override func setUp() async throws {
       promise = Promise()
       module = ReactNativeTracerProviderModule()
-      // Flush any pending exports from previous test
-      _ = ReactNativeTracerProviderTests.exporter.flush(explicitTimeout: 5.0)
 
       // Initial wait to let Embrace SDK initialize (especially important for first test)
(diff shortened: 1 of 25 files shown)


packages/react-native-tracer-provider/test-project/ios/RNEmbraceTracerProviderTests/RNEmbraceTracerProviderTests.swift
```diff
@@ -87,8 +87,6 @@ class ReactNativeTracerProviderTests: XCTestCase {
   override func setUp() async throws {
       promise = Promise()
       module = ReactNativeTracerProviderModule()
-      // Flush any pending exports from previous test
-      _ = ReactNativeTracerProviderTests.exporter.flush(explicitTimeout: 5.0)
```
hamudi-ahmed, line 91: The latest OpenTelemetry Swift Sdk release introduced async method overloads for `SpanExporter`, which caused a compilation error here as Swift's overload resolution silently switched to the async version (since `setUp()` is async)

I've just removed the offending line as this is dead code anyway (`TestExporter.flush` is a noop)

jpmunz (approved):

jpmunz:
@hamudi-ahmed code change lgtm, though looks like there are a couple remaining check failures as a result

hamudi-ahmed:
Checks are all passing now. The previous dependency review failure was a false positive and passed on a subsequent re-run - the dependency it was flagging was actually the fixed version (joi@17.13.4), but this hadn't propagated yet as it had only just been released.

PR #891, opened 2026-06-19
[EMBR-12633] Update core package for android sdk v8.4.0

hamudi-ahmed:
## Goal

Updates the core package for Android SDK v8.4.0:

- Replaces deprecated `Embrace.getInstance()` with `Embrace.INSTANCE` in `EmbraceManagerModule`
- Updates unit tests:
	- Switched test code to use the OTel Kotlin API
	- Tests now run using Roboelectric + JUnit 4 (required to start the Android SDK)

## Testing

Covered by the core Android unit tests (now passing) - the remaining packages will be updated in subsequent PRs


packages/core/test-project/android/app/src/test/java/io/embrace/rnembracecoretest/RNEmbraceCoreTest.kt
```diff
@@ -341,78 +289,78 @@ class RNEmbraceCoreTest {
...
             val logWithNoSeverity = allValues[6].asSequence().withIndex().elementAt(0).value
             assertEquals("ERROR", logWithNoSeverity.severityText)
-            assertEquals("a message without severity", logWithNoSeverity.body.asString())
-            assertEquals("sys.log", logWithNoSeverity.attributes.get(AttributeKey.stringKey("emb.type")))
-            assertEquals(4, logWithNoSeverity.attributes.size())
-            assertNotNull(logWithNoSeverity.attributes.get(AttributeKey.stringKey("log.record.uid")))
-            assertEquals("stacktrace as string", logWithNoSeverity.attributes.get(AttributeKey.stringKey("emb.stacktrace.rn")))
+            assertEquals("a message without severity", logWithNoSeverity.body.toString())
+            assertEquals("sys.log", logWithNoSeverity.attributes["emb.type"])
+            assertEquals(6, logWithNoSeverity.attributes.size)
```
jpmunz, line 379: looks like these consistently jumped by 2, do we know what the additional attributes now included are?

packages/core/package.json
```diff
@@ -59,7 +59,7 @@
   "scripts": {
     "js:shared": "mkdir -p ./src/interfaces/common && cp -r ../../common/js/* ./src/interfaces/common",
     "ios:shared": "mkdir -p ./ios/common && cp -r ../../common/ios/* ./ios/common/",
-    "android:test": "cd test-project && yarn install && cd android && ./gradlew :app:detek && ./gradlew :app:testDebugUnitTest && cd ../../",
+    "android:test": "cd test-project && yarn install && cd android && ./gradlew :app:detekt && ./gradlew :app:testDebugUnitTest && cd ../../",
```
jpmunz, line 62: was this just a typo previously?

jpmunz (commented):
Some additional comments:
* we also use `getInstance` in the install scripts and expo plugin
* we refer to it in the RN docs as well https://embrace.io/docs/react-native/integration/session-reporting/?platform=android, might be worth starting a draft PR in that repo to keep track of the changes we'll need documented for the bump
* shouldn't the Android unit tests in packages/react-native-tracer-provider and packages/react-native-otlp still be failing since this only updated core?

packages/core/test-project/android/app/src/test/java/io/embrace/rnembracecoretest/RNEmbraceCoreTest.kt
hamudi-ahmed: We do - in 8.4.0 there are now two additional `emb.state.*` attributes - `emb.state.network` and `emb.state.screen-automatic`. There's also a pre-existing `emb.state` attribute (foreground/background state), which we're also not asserting on. 

I'm in two minds about whether we should add assertions for these attributes - they're not really related to the `logMessage` call that's being tested, but without them the count assertion is a bit confusing...

packages/core/package.json
hamudi-ahmed: yeah it's a typo. It still works because gradle has task name abbreviation which means you don't have to use the full name when invoking a task

hamudi-ahmed:
> * we also use `getInstance` in the install scripts and expo plugin
> * we refer to it in the RN docs as well [embrace.io/docs/react-native/integration/session-reporting?platform=android](https://embrace.io/docs/react-native/integration/session-reporting/?platform=android), might be worth starting a draft PR in that repo to keep track of the changes we'll need documented for the bump

good catch 👍🏾  - I've updated the scripts and expo plugin, will create a docs PR


> * shouldn't the Android unit tests in packages/react-native-tracer-provider and packages/react-native-otlp still be failing since this only updated core?

CI runs a change detection script, so only the core tests were executed for this PR

packages/core/test-project/android/app/src/test/java/io/embrace/rnembracecoretest/RNEmbraceCoreTest.kt
jpmunz: ya good point, I don't have a strong preference which but I think we should go either:
a) remove the assertion on the attributes.size altogether, e.g. we don't care to assert on the unrelated attributes
b) OR, adding in the full set for consistency

hamudi-ahmed: agreed - as discussed IRL, I've added the full set for consistency for now, and we can think about removing these in future once we have equivalent coverage in the integration tests

jpmunz (approved):

PR #899, opened 2026-06-23
feat(plugin): wire custom OTLP exporters into native iOS Embrace start

### LINKED ISSUE #848: iOS: custom OTLP exporter emits nothing — no POST sent, even for explicitly-created tracer-provider spans
**Summary**

  On iOS, the @embrace-io/react-native-otlp custom exporter configured via useEmbrace({ exporters }) never sends any OTLP request to the
  configured endpoint. Embrace's own backend receives the session (visible in the Embrace dashboard), but the custom log/trace exporters emit
  zero network requests — verified against both a local HTTP catcher and a remote HTTPS OTLP endpoint, in Debug (simulator) and Release
  (physical device), for logs and traces.

  This appears related to #653, but is a cleaner/stronger reproduction: that report at least sees HTTP 200 POSTs; here no POST is emitted at
  all, including for spans explicitly created via @embrace-io/react-native-tracer-provider.

  **Environment**

  - @embrace-io/react-native: 6.6.0 (native iOS SDK 6.16.2)
  - @embrace-io/react-native-otlp: 6.6.0
  - @embrace-io/react-native-tracer-provider: 6.6.0
  - react-native: 0.81.5, Expo: 54 (managed, config plugin), Hermes, New Architecture enabled
  - Platform: iOS (simulator iOS 26 + physical device); package manager: pnpm

  **What I expected**

  With exporters configured, Embrace forwards telemetry (and at minimum any explicitly-created OTel spans) to the configured OTLP endpoint
  over HTTP.

  **What happens**

  - useEmbrace(...) reports isStarted: true.
  - A test span is created on start (see below) and span.end() is called — confirmed via JS log.
  - No OTLP request is ever made: a local catcher receives 0 requests; a remote OTLP backend receives nothing.
  - The same app's session data does reach Embrace's own backend normally.

  **Configuration (JS)**

  ```
const { isStarted } = useEmbrace({
    trackUnhandledRejections: true,
    ios: { disabledUrlPatterns: ["<otlp-host>"] },
    exporters: {
      logExporter:   { endpoint: "https://<host>/v1/logs",   headers: [{ key: "apikey", token: "<token>" }], timeout: 30000 },
      traceExporter: { endpoint: "https://<host>/v1/traces", headers: [{ key: "apikey", token: "<token>" }], timeout: 30000 },
    },
  });

  // also tried with an explicit span:
  const { tracerProvider } = useEmbraceNativeTracerProvider({}, isStarted);
  useEffect(() => {
    if (!isStarted || !tracerProvider) return;
    const span = tracerProvider.getTracer("otlp-test").startSpan("otlp-test-span");
    span.end();
  }, [isStarted, tracerProvider]);
```
  App IDs + apiToken are injected via the Expo config plugin (Embrace-Info.plist); native SDK starts fine.

  **What we ruled out**

  - **Endpoint/auth/routes are fully working** — confirmed end-to-end, not just by status code. A direct OTLP POST (curl) with the same apikey
  header to both …/v1/logs and …/v1/traces returns HTTP 200 {"partialSuccess":{}}, and the posted records are subsequently queryable in the
  backend (a smoke-test log and a smoke-test span, both found by service.name). Both JSON and protobuf bodies are accepted. So the routes
  ingest correctly; the app simply never POSTs to them.
  - **Cleartext/ATS** — also reproduced against an HTTPS endpoint, so it's not ATS blocking http://localhost.
  - **No telemetry to export** — reproduced with an explicit tracer-provider span (span.end() confirmed in logs).
  - **Build type** — reproduced in Debug (simulator) and Release (physical device).
  - **disabledUrlPatterns** — set per the OTLP docs; no change.
  - **The OTLP native pod** (RNEmbraceOTLP) and tracer-provider pod (RNEmbraceTracerProvider) are confirmed compiled into the build.

  **Questions**

  1. On iOS, what actually drives/flushes the custom OTLP exporters, and what telemetry is routed through them (session spans? only
  tracer-provider spans? logs)?
  2. Is there any required native setup beyond useEmbrace({ exporters }) for the iOS exporter to emit?
  3. Is there a way to enable verbose SDK logging to see the exporter attempting (or skipping) a flush?

### PR DESCRIPTION
**Pull request by giladAiola**
## Problem

Closes #848. (Related: #653.)

On **Expo iOS**, the config plugin starts Embrace **natively** at launch (the generated `EmbraceInitializer.start()`), which runs before — and pre-empts — the JS `useEmbrace({ exporters })` path. As a result, custom OTLP exporters configured in JS are never attached to the iOS SDK: no export POST is ever sent, even for explicitly-created tracer-provider spans (#848). The native start is intentional (early native crash capture), so JS-supplied exporters can't be wired after the fact.

## Fix

Add an optional **`iOSExporters`** plugin prop (typed with the existing `OTLPExporterConfig` — the same `traceExporter` / `logExporter` shape used by `useEmbrace`). When provided, the plugin generates an `EmbraceInitializer.swift` that:

- builds `OtlpHttpTraceExporter` / `OtlpHttpLogExporter` from the config (endpoint, headers, timeout) and passes them via `Embrace.Options(export:)`, so exporters take effect **while keeping native crash capture**;
- adds the OTLP endpoint host(s) to `URLSessionCaptureService.Options(ignoredURLs:)` — the native equivalent of the JS `iOSConfig.disabledUrlPatterns` — so the exporter's own upload POSTs (`/v1/traces`, `/v1/logs`) aren't instrumented into spans, which would otherwise feed back into the exporter.

When `iOSExporters` is omitted the generated initializer is unchanged (minimal `Embrace.Options(appId:platform:)`), so this is fully backward-compatible.

## Status

**This is a working fix, currently deployed in production in our app.** We run it on Expo iOS via a fork of this SDK; with `iOSExporters` wired into the native start, OTLP traces and logs reach our collector reliably, and the `ignoredURLs` exclusion stops the exporter's own upload requests from being recaptured.

## Notes

- iOS-only — the native-start pre-emption is iOS-specific; Android wires exporters through its own path.
- The generated Swift imports `OpenTelemetrySdk`, `OpenTelemetryProtocolExporterCommon`, and `OpenTelemetryProtocolExporterHttp` (pulled in transitively by EmbraceIO's OTLP setup). Happy to adjust the module surface if you'd prefer.
- Happy to add a unit test asserting the generated initializer string if that fits your conventions.

### DIFF
--- a/packages/core/src/plugin/types.ts
+++ b/packages/core/src/plugin/types.ts
@@ -1,4 +1,4 @@
-import {OTLPExporterConfig} from "../interfaces";
+import {IOSConfig, OTLPExporterConfig} from "../interfaces";
 
 type EmbraceProps = {
   /**
@@ -30,15 +30,18 @@ type EmbraceProps = {
   productModuleName?: string;
 
   /**
-   * iOSExporters configures custom OTLP exporters that are wired into the *native* Embrace start on iOS
-   * (in the generated `EmbraceInitializer.swift`). This is required because the Expo plugin starts Embrace
-   * natively at launch (for early crash capture), which pre-empts the JS `useEmbrace({exporters})` path — so
-   * on Expo iOS, JS-supplied exporters are never attached (see issue #653). Providing them here injects them
-   * at the native start instead, keeping both crash capture and OTLP forwarding. The OTLP endpoint hosts are
-   * also excluded from URLSession capture so the exporter's own upload requests aren't instrumented (the
-   * native equivalent of `iOSConfig.disabledUrlPatterns`). Values are baked in at prebuild.
+   * iOSExporters configures custom OTLP exporters that are wired into the native Embrace start on iOS
+   * (in the generated `EmbraceInitializer.swift`).
    */
   iOSExporters?: OTLPExporterConfig;
+
+  /**
+   * iOSConfig mirrors the iOS-specific options from the JS `useEmbrace({ios})` path for the native
+   * start (the generated `EmbraceInitializer.swift`). Only applied alongside `iOSExporters`, since
+   * that's the only path that generates the native start. Currently `disableNetworkSpanForwarding`
+   * and `disabledUrlPatterns` are honored.
+   */
+  iOSConfig?: IOSConfig;
 };
 
 export {EmbraceProps};
--- a/packages/core/src/plugin/types.ts
+++ b/packages/core/src/plugin/types.ts
@@ -1,3 +1,5 @@
+import {OTLPExporterConfig} from "../interfaces";
+
 type EmbraceProps = {
   /**
    * androidAppId is your Embrace App ID for Android.
@@ -26,6 +28,17 @@ type EmbraceProps = {
    * for more details on why this is required.
    */
   productModuleName?: string;
+
+  /**
+   * iOSExporters configures custom OTLP exporters that are wired into the *native* Embrace start on iOS
+   * (in the generated `EmbraceInitializer.swift`). This is required because the Expo plugin starts Embrace
+   * natively at launch (for early crash capture), which pre-empts the JS `useEmbrace({exporters})` path — so
+   * on Expo iOS, JS-supplied exporters are never attached (see issue #653). Providing them here injects them
+   * at the native start instead, keeping both crash capture and OTLP forwarding. The OTLP endpoint hosts are
+   * also excluded from URLSession capture so the exporter's own upload requests aren't instrumented (the
+   * native equivalent of `iOSConfig.disabledUrlPatterns`). Values are baked in at prebuild.
+   */
+  iOSExporters?: OTLPExporterConfig;
 };
 
 export {EmbraceProps};
--- a/packages/core/src/plugin/withIosEmbrace.ts
+++ b/packages/core/src/plugin/withIosEmbrace.ts
@@ -13,6 +13,7 @@ import {
   updateBuildProperty,
 } from "./xcodeproj";
 import {EmbraceProps} from "./types";
+import {ExporterConfig, OTLPExporterConfig} from "../interfaces";
 import {addAfter, hasMatch} from "./textUtils";
 import {writeIfNotExists} from "./fileUtils";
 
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: packages/core/src/plugin/types.ts [hunk 1]
```diff
@@ -26,6 +28,17 @@ type EmbraceProps = {
...
   productModuleName?: string;
+
+  /**
+   * iOSExporters configures custom OTLP exporters that are wired into the *native* Embrace start on iOS
+   * (in the generated `EmbraceInitializer.swift`). This is required because the Expo plugin starts Embrace
+   * natively at launch (for early crash capture), which pre-empts the JS `useEmbrace({exporters})` path — so
+   * on Expo iOS, JS-supplied exporters are never attached (see issue #653). Providing them here injects them
+   * at the native start instead, keeping both crash capture and OTLP forwarding. The OTLP endpoint hosts are
+   * also excluded from URLSession capture so the exporter's own upload requests aren't instrumented (the
+   * native equivalent of `iOSConfig.disabledUrlPatterns`). Values are baked in at prebuild.
```
hamudi-ahmed: nit: I think these comments could be simplified
```suggestion
   * iOSExporters configures custom OTLP exporters that are wired into the native Embrace start on iOS
   * (in the generated `EmbraceInitializer.swift`).
```

packages/core/src/plugin/withIosEmbrace.ts
```diff
@@ -67,7 +68,132 @@ const withIosEmbraceAddKSCrashPod: ConfigPlugin = expoConfig => {
...
+            endpoint: URL(string: "${escapeSwift(exporter.endpoint)}")!,
+            config: OtlpConfiguration(timeout: ${timeoutExpr}, headers: ${varName}Headers),
+            httpClient: BaseHTTPClient(session: URLSession(configuration: ${varName}URLConfig)),
+            envVarHeaders: ${varName}Headers
+        )`;
+};
+
+// Initializer that wires custom OTLP exporters into the *native* Embrace start. The native start
+// pre-empts the JS `useEmbrace({exporters})` path on Expo iOS, so exporters must be attached here to
+// take effect (while keeping native crash capture). See issue #653.
```
hamudi-ahmed: nit:
```suggestion
// Initializer that wires custom OTLP exporters into the *native* Embrace start.
```

packages/core/src/plugin/withIosEmbrace.ts
```diff
@@ -67,7 +68,132 @@ const withIosEmbraceAddKSCrashPod: ConfigPlugin = expoConfig => {
...
+  return `import Foundation
+import EmbraceIO
+import OpenTelemetrySdk
+import OpenTelemetryProtocolExporterCommon
+import OpenTelemetryProtocolExporterHttp
+
+@objcMembers class EmbraceInitializer: NSObject {
+    // Start the EmbraceSDK natively WITH custom OTLP exporters. Wiring exporters here (rather than via
+    // JS useEmbrace) is required on Expo iOS because the native start runs at launch and pre-empts the
+    // JS start. See https://github.com/embrace-io/embrace-react-native-sdk/issues/653
```
hamudi-ahmed: nit:
```suggestion
    // Start the EmbraceSDK natively WITH custom OTLP exporters.
```

packages/core/src/plugin/withIosEmbrace.ts
```diff
@@ -67,7 +68,132 @@ const withIosEmbraceAddKSCrashPod: ConfigPlugin = expoConfig => {
...
+    static func start() -> Void {
+${traceSetup}
+${logSetup}
+        let export = OpenTelemetryExport(spanExporter: ${spanArg}, logExporter: ${logArg})
+
+        let servicesBuilder = CaptureServiceBuilder()
+        servicesBuilder.add(.urlSession(options: URLSessionCaptureService.Options(
+            injectTracingHeader: true,
```
hamudi-ahmed: I think it would be better to allow this to be configured via an explicit `disableNetworkSpanForwarding` option in the config plugin.

packages/core/src/plugin/withIosEmbrace.ts
```diff
@@ -67,7 +68,132 @@ const withIosEmbraceAddKSCrashPod: ConfigPlugin = expoConfig => {
...
+            requestsDataSource: nil,
+            ignoredURLs: [${ignoredURLsLiteral}]
```
hamudi-ahmed, line 150: Same as above - I think we should allow this to be configured via an explicit `disabledUrlPatterns` option in the config plugin instead of inferring from the exporter endpoints

packages/core/src/plugin/withIosEmbrace.ts
```diff
@@ -67,7 +68,132 @@ const withIosEmbraceAddKSCrashPod: ConfigPlugin = expoConfig => {
...
+  );
+  const ignoredURLsLiteral = ignoredHosts
+    .map(h => `"${escapeSwift(h)}"`)
+    .join(", ");
+
+  return `import Foundation
+import EmbraceIO
+import OpenTelemetrySdk
+import OpenTelemetryProtocolExporterCommon
+import OpenTelemetryProtocolExporterHttp
```
hamudi-ahmed, line 137: Not something that needs to be resolved in this PR, but we'll need to think about before shipping: these dependencies are pulled in via the `react-native-otlp` package but the expo plugin is shipped with the `core` package, which may be confusing.

We could either have the expo plugin check that the package is installed and if not log a warning/noop, or have the plugin add the required native dependencies itself (effectively replacing the `react-native-otlp` package)

hamudi-ahmed:
Hi @giladAiola, thanks for the contribution! This mostly looks good, and plugs a real gap with the otlp package - I think wiring the exporters in at the native start is the right approach, and it lines up well with our plans for the `react-native-otlp` package longer term.

We'd like to accept it, with a few next steps:

- We'd like to merge this into a feature branch rather than main, so we can build the equivalent Android implementation on top and ship the two together. I've created a branch for this called `feature/expo-plugin-otlp-exporters`  - please update the base branch on the PR accordingly.
- I've added a few comments/suggestions - if you're able to address some of these that's great, but since it's going into a feature branch I don't think they're required to merge this. We can address those later as follow-ups if needs be.
- Similarly, if you'd like to add unit tests for the generated initializer that would be great, but we can also add those later.
- Please read and sign the Embrace [CLA](https://docs.google.com/forms/d/e/1FAIpQLSct_OV9j5-yGCXkhMKOUKpwTxFWdwCQYTPeESk39lOM6k3uKA/viewform) - this is required before we can approve and merge the PR.


Thanks again for pushing this forward!

giladAiola:
Thanks for the thorough review @hamudi-ahmed! Pushed a couple of follow-up commits addressing the feedback:

**Explicit config instead of inference** (`disabledUrlPatterns` / `disableNetworkSpanForwarding`)
Replaced the implicit "infer ignored hosts from the exporter endpoints" logic with an explicit `iOSConfig` plugin prop, reusing the existing `IOSConfig` type from `../interfaces`:
- `iOSConfig.disabledUrlPatterns` → `URLSessionCaptureService.Options(ignoredURLs:)`
- `iOSConfig.disableNetworkSpanForwarding` → `URLSessionCaptureService.Options(injectTracingHeader:)`

This matches how `disabledUrlPatterns` / `disableNetworkSpanForwarding` already behave in the JS `useEmbrace({ ios })` path, so the config surface is consistent across JS and the native start.

One thing to flag: with the inference gone, anyone wiring exporters now has to list their OTLP endpoint(s) in `disabledUrlPatterns` themselves to avoid the exporter's own upload POSTs being recaptured into spans (the feedback loop). That's the same contract as JS, just wanted to call it out since it's a slightly sharper edge than the old auto-exclusion. Happy to add a note in the docs/README if useful.

**Nits** — applied all three (simplified the `iOSExporters` doc comment and the two generated-initializer comments).

**Unit tests** — added tests for `withIosEmbraceAddInitializer` on the OTLP path: a full assertion of the generated `EmbraceInitializer.swift` against a fixture (trace + log exporters, custom header/timeout), log-only configuration, explicit `disabledUrlPatterns`, default-empty `ignoredURLs`, and `disableNetworkSpanForwarding`.

**Base branch** — updated the PR base to `feature/expo-plugin-otlp-exporters` as requested.

**`react-native-otlp` deps vs `core` package** (your comment on the OTLP imports) — agreed this needs sorting before shipping, but as you suggested I've left it out of this PR; happy to pick it up as a follow-up on the feature branch (lean toward having the plugin detect the package and warn/noop if it's missing).

CLA signed. 🙏

hamudi-ahmed (approved):
LGTM - thanks for taking the time to add tests and address all the comments! 🙌🏾 

I think the failing checks are due to unrelated tooling/CI issues that are either fixed or soon-to-be fixed upstream, so can safely be ignored here - happy to merge this as-is.

I don't have an ETA for when the Android changes will be scheduled just yet, but I'll be sure to keep you posted with any updates.

Thanks again for the contribution @giladAiola!

PR #912, opened 2026-07-02
[EMBR-12845] Fix OTel Java dependency resolution issues

hamudi-ahmed:
## Goal

Currently in `react-native-otlp` we are pinning all the otel-java dependencies to 1.51.0, and also require customers do the same, in order to work around a kotlin metadata version conflict. In the [docs](https://embrace.io/docs/react-native/changelog/#621) we state that this is only necessary if you’re using Kotlin 1.x, but while testing against Android v8 (using Kotlin 2.0) I discovered that isn’t the case and compilation still fails. 

Forcing 1.51.0 just masks the real problem - opentelemetry-exporter-otlp 1.52.0+ pulls in okhttp 5, which carries kotlin 2.2 metadata, and this is what causes builds to fail.

React Native ships with okhttp 4, so a much better solution is to exclude okhttp when declaring the opentelemetry dependency so that the app’s resolved okhttp version is used instead.

This is what opentelemetry-java recommend doing, and they have confirmed that the exporters will continue to work fine with okhttp 4: https://github.com/open-telemetry/opentelemetry-java/issues/8001#issuecomment-3787675805

Doing this means that folks no longer need to add extra gradle config to force the otel dependencies to an old version, and also means we can bump our dependencies to align with the Android SDK (1.62.0).

## Testing

Tested manually with the integration test templates - confirmed that all the templates compile and run, and that logs and traces are exported as expected.


jpmunz (approved):
lgtm, we should also update the docs if this workaround is no longer necessary

PR #922, opened 2026-07-07
[EMBR-12894] Split integration tests into per-platform build/test jobs

hamudi-ahmed:
## Goal

Restructures the integration-test workflow so tests run more reliably on BrowserStack while keeping builds as fast as possible. 

Previously we were firing all test runs in parallel (or whenever the build step succeeded) and relying on wdio retrying requests + the `onWorkerStart` hook to mitigate the limited number of available browserstack slots. Despite that mitigation, we've been seeing lots of test runs failing due to session creation requests timing out and `BROWSERSTACK_QUEUE_SIZE_EXCEEDED` errors.

It looks like the situation got worse after the upgrade to wdio v9, which no longer retries timeouts in session creation requests. After `connectionRetryTimeout` elapses wdio just aborts and errors out, so it no longer recovers by itself.

The fix caps the number of in-flight sessions upfront to what our plan can handle, and decouples test runs from the build phase so that tests can be retried without having to re-build the test fixture. 

This greatly reduces the likelihood of tests failing due to request timeouts and queue size limits, and since BS slots are limited anyway it doesn't really affect overall test run times - it just shifts the bottleneck to somewhere we can better control. If the number of slots is ever increased, we have a clear knob to turn to increase concurrency (`max-parallel` in the workflow and `maxInstances` in the wdio config).

## Workflow (`integration-test.yml`)
- Split the single build+test job into separate **build** and **test** jobs, per platform:
  `build-android`, `build-ios`, `test-android`, `test-ios`.
- Builds run fully in parallel as before and upload the test app as an artifact that the test jobs then download. This keeps fixture builds fast and isolates the bottleneck to the actual test jobs.
- Each test job is `max-parallel: 1`, which keeps max 2 sessions in flight (1 per platform)
- Added a single `gate` job that computes the run condition once (per platform), removing the
  duplicated `if:` across jobs.
- Added a `platform` input (`both` | `android` | `ios`) to the manual trigger to support manual runs for a single platform. All other triggers run both platforms.

## wdio config (`browserstack.conf.ts`)
- Reduced `maxInstances` to 1 since we can only run 1 session at a time anyway.
- Upped `connectionRetryTimeout` to 15 mins, matching BrowserStack's 15-min server-side queue hold,
  so a queued session isn't aborted early (wdio v9 makes this a hard, non-retried abort).
- Reduced `specFileRetriesDelay` to 5s - we shouldn't see `BROWSERSTACK_QUEUE_SIZE_EXCEEDED` errors any more, and it just slowed test runs down unnecessarily (turns out wdio now applies this as a pre-start sleep before every worker, not just when a spec file fails).
- Removed the queue-check jitter (unnecessary now that workers run sequentially and max 2 sessions are ever in flight).

## Testing

Tested with a manual workflow trigger

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -21,11 +21,12 @@ permissions:
   pull-requests: read
 
 jobs:
-  gate:
+  prepare:
     runs-on: ubuntu-latest
     outputs:
       android: ${{ steps.decide.outputs.android }}
       ios: ${{ steps.decide.outputs.ios }}
+      matrix: ${{ steps.decide.outputs.matrix }}
     steps:
       - id: decide
         env:
@@ -42,14 +43,15 @@ jobs:
           fi
           echo "android=$android" >> "$GITHUB_OUTPUT"
           echo "ios=$ios" >> "$GITHUB_OUTPUT"
+          echo 'matrix=["expo-rn74","rn82","rn81","rn80","rn79","rn78","rn77","rn76","rn75","rn74","rn73","rn72","rn71"]' >> "$GITHUB_OUTPUT"
 
   build-android:
-    needs: gate
-    if: ${{ needs.gate.outputs.android == 'true' }}
+    needs: prepare
+    if: ${{ needs.prepare.outputs.android == 'true' }}
     strategy:
       fail-fast: false
       matrix:
-        test-app: [expo-rn74, rn82, rn81, rn80, rn79, rn78, rn77, rn76, rn75, rn74, rn73, rn72, rn71]
+        test-app: ${{ fromJSON(needs.prepare.outputs.matrix) }}
 
     runs-on: ubuntu-latest
 
@@ -92,12 +94,12 @@ jobs:
           retention-days: 1
 
   build-ios:
-    needs: gate
-    if: ${{ needs.gate.outputs.ios == 'true' }}
+    needs: prepare
+    if: ${{ needs.prepare.outputs.ios == 'true' }}
     strategy:
       fail-fast: false
       matrix:
-        test-app: [expo-rn74, rn82, rn81, rn80, rn79, rn78, rn77, rn76, rn75, rn74, rn73, rn72, rn71]
+        test-app: ${{ fromJSON(needs.prepare.outputs.matrix) }}
 
     runs-on: macos-26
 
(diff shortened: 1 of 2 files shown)


.github/workflows/integration-test.yml
```diff
@@ -66,7 +135,91 @@ jobs:
...
+          retention-days: 1
+
+  test-android:
+    needs: [gate, build-android]
+    if: ${{ !cancelled() && needs.gate.outputs.android == 'true' }}
+    strategy:
+      fail-fast: false
+      max-parallel: 1
+      matrix:
+        test-app: [expo-rn74, rn82, rn81, rn80, rn79, rn78, rn77, rn76, rn75, rn74, rn73, rn72, rn71]
```
jpmunz: would there be a way to share the matrix across each step that needs it?

integration-tests/browserstack.conf.ts
```diff
@@ -87,7 +88,8 @@ exports.config = {
   waitforTimeout: 10000,
   // wdio v9 enforces this as a hard abort on session creation (got->fetch); a
   // timed-out newSession is no longer retried, so give BrowserStack room to allocate.
-  connectionRetryTimeout: 300000,
+  // 15 minutes to match browserstack's session timeout - this is the max time a session can be queued before it is aborted.
+  connectionRetryTimeout: 900000,
```
jpmunz, line 92: I'm maybe missing how these numbers interact with each other, is it possible to get this high given `connectionRetryCount` is 3 and `waitforTimeout` is 10000?

hamudi-ahmed: Yeah, the config option naming is confusing - I went down a bit of a rabbit hole with this but glossed over it in the PR...

`connectionRetryTimeout` is the client-side timeout for a single webdriver HTTP request. When we're queued, the `POST /session` request is just held open until a slot frees. It's unlikely to reach 15 mins now as tests only take a few mins to run, but once we start running more spec files that could increase. I set it to match BrowserStack's own server-side limit so we don't give up before they would.

Annoyingly, `connectionRetryCount` only retries retryable failures (connection resets, 5xx, etc.). In wdio v9 a `connectionRetryTimeout` gets applied as an `AbortSignal.timeout` on the request, and surfaces as a `TimeoutError`, which isn't in wdio's retryable set — so a timed-out new session request isn't retried at all, it's just a hard fail. You can see in the CI logs that these timeouts are never retried. From what I can tell this is a subtle change in behaviour in wdio v9 but I'm not sure if it's a bug or by design.
 
`waitforTimeout` is completely unrelated - it only affects test commands that start with `waitFor`.

.github/workflows/integration-test.yml
hamudi-ahmed: Yeah, good idea. I've moved it up into the `gate` step and renamed that to `prepare`. I think that's probably fine for now, but if we end up needing a more complex matrix in the future we could move it out into a JSON file.

integration-tests/browserstack.conf.ts
jpmunz: ah ok, thanks for the explanation

jpmunz (approved):

PR #946, opened 2026-07-23
[EMBR-13241] Reinstate and uplift Integration tests

hamudi-ahmed:
## Goal

Reinstates the failing integration test specs (currently locally only) and re-works the test suite around a small set of shared helpers and golden-file comparisons. 

Specs now assert against real payloads and are stable across runs and platforms.

This is a first pass at improving the test suite, and doesn't add any new test cases - the plan is to extend this to add more test cases (logs, network requests, etc), and then look at getting tests running in CI with browserstack + remote mock api.

Key changes:
- Payload normalization: a single normalizePayloads step sorts captured spans/logs into meaningful buckets (session, view, perf, network, snapshots, logs) and hides Android/iOS differences behind one shape.

- Golden-file assertions: committed per-platform golden captures for each scenario, compared via a toMatchGoldenFile matcher. Volatile fields (ids, timestamps, session.id) are checked for presence, not value, so runs stay stable.

- Custom matchers: readable expect extensions (toMatchGoldenFile, toMatchSpan, toHaveAttributes, toHaveParentSpan, toHaveSpanNames, etc.) replace manual attribute lookups in specs - this makes error output on failures much clearer.

- Payload source abstraction: specs read captured payloads through a PayloadSource interface - currently there's only a local mockserver implementation, but the plan is to extend this to the remote mock API on browserstack in future.

- Rewritten specs — session, navigation, and tracer-provider specs now use the new helpers and matchers, and read more clearly.

- Config/cleanup — added the spec reporter for readable output, updated the wdio config, simplified the session helper (background via mobile: backgroundApp), and removed the old hand-rolled span parser.

## Testing

Confirmed that all the current spec files pass on both Android and iOS when running locally 🎉


integration-tests/specs/tracer_provider.test.ts
```diff
@@ -1,283 +1,44 @@
...
-    }
+    const p = await source.getPayloads();
+    expect(p.perfSpans).toMatchGoldenFile("tracer-test-spans", "perfSpans");
+    expect(p.spanSnapshots).toMatchGoldenFile("tracer-test-spans", "spanSnapshots");
   });
 
-  it("should record nested spans", async () => {
-    const generateTestSpans = await driver.$("~GENERATE NESTED SPANS");
-    await generateTestSpans.click();
+  it("records nested spans with correct parent relationships", async () => {
```
jpmunz, line 37: this case maybe isn't needed since it checks the same golden as "records test spans and an unfinished-span snapshot"

integration-tests/specs/navigation.test.ts
```diff
@@ -1,161 +1,44 @@
...
+import {getPayloadSource} from "../helpers/payload_source";
+import {endSession} from "../helpers/session";
 import {EmbraceSpanData} from "../typings/embrace";
-import {
-  commonEmbraceSpanAttributes,
-  embraceSpanDefaults,
-  sortSpanAttributes,
-} from "../helpers/span";
+
+// On iOS backgrounding surfaces as the transitional "inactive" state (see EMISSION-MODEL.md).
```
jpmunz: doesn't look like this EMISSION-MODEL.md file exists anymore

integration-tests/specs/navigation.test.ts
```diff
@@ -1,161 +1,44 @@
...
-    // On iOS we get the "inactive" state which represents a transition between foreground and background instead of
-    // "background" due to how our integration tests run
-    // See https://reactnative.dev/docs/appstate#app-states
-    return driver.isAndroid ? "background" : "inactive";
-  };
+  const payloadSource = getPayloadSource();
+
+  // Establish a known starting screen so the spec is independent of run order.
+  beforeEach(async () => {
+    await driver.$("~LOG TESTING").click();
```
jpmunz, line 19: not for this PR but may be good to have a couple helpers for this pattern we have in a few spots of clicking and then waiting, or clicking and then ending the session

integration-tests/helpers/compare.ts
```diff
@@ -0,0 +1,209 @@
...
+        errors.push(`missing volatile attribute "${key}"`);
+      }
+    } else if (!actualMap.has(key)) {
+      errors.push(`missing attribute "${key}"`);
+    } else if (actualMap.get(key) !== value) {
+      errors.push(`attribute "${key}" expected "${value}", got "${actualMap.get(key)}"`);
+    }
+  }
+  for (const key of actualMap.keys()) {
+    if (!expectedMap.has(key) && !VOLATILE_ATTR_KEYS.has(key)) {
```
jpmunz: since volatile keys are still checked for presence should ` && !VOLATILE_ATTR_KEYS.has(key)` be removed here?

integration-tests/helpers/normalize.ts
```diff
@@ -0,0 +1,82 @@
...
+    spanSnapshots: [],
+    logs: [],
+    ignored: [],
+  };
+
+  spanEnvelopes.forEach(env => {
+    (env.data?.spans ?? []).forEach(span => {
+      if (span.name === "emb-session") {
+        out.sessionSpans.push(span);
+      } else if (getAttribute(span, "emb.type") === "ux.view") {
```
jpmunz: nit, this might be common enough to warrant a `getEmbType` helper

integration-tests/wdio.conf.ts
```diff
@@ -180,7 +178,15 @@ export const config: Options.Testrunner = {
   // Test reporter for stdout.
   // The only one supported by default is 'dot'
   // see also: https://webdriver.io/docs/dot-reporter
-  reporters: ["json"],
+  reporters: [
+    [
+      "spec",
+      {
+        realTimeReporting: true,
+
```
jpmunz:
```suggestion
```

jpmunz (approved):
looks great! Nice how clear the specs themselves got, just a couple minor comments but nothing blocking

integration-tests/specs/navigation.test.ts
hamudi-ahmed: good idea - I'll make a note to add that as we add more specs. I think wdio supports custom commands which might be a nice way to do that

integration-tests/specs/tracer_provider.test.ts
hamudi-ahmed: this one is actually checking against a different golden (tracer-nested-spans) that exercises the various ways of setting span parentage (generateNestedSpans vs generateTestSpans in the test harness). Both of these tests are a bit coarse though, I think we could probably split both out into more fine-grained tests/assertions in a future iteration

integration-tests/helpers/compare.ts
hamudi-ahmed: yeah, good catch

PR #950, opened 2026-07-27
[EMBR-6511] Remove unused test app

hamudi-ahmed:
## Goal

Removes the now unused basic-test-app (superseded by integration test templates)


scripts/run-clang-format.sh
```diff
@@ -12,7 +12,6 @@ fi
 
 # 1) Paths to formart
 SOURCE_DIRS=(
-    "$(pwd)/integration-tests/basic-test-app/ios/basictestapp"
```
hamudi-ahmed, line 15: Removed the reference to basic-test-app here, but looks like this script isn't currently used, and doesn't actually work any more - I've created a separate ticket to review/fix it

jpmunz (approved):

PR #958, opened 2026-07-29
[EMBR-4922] Expand integration test specs

hamudi-ahmed:
## Goal

Broadens the integration suite from session/navigation/tracer coverage to the rest of the test harness screens.

- Added new specs for the log, network, user and redux screens
- Expanded the session, navigation and tracer specs with additional test cases
- Added goldens for the new scenarios, and refreshed/fixed some existing ones
- Updated helpers:  renamed `session.ts` → `app.ts` and added a shared `tap` helper, plus a log comparator and a `toHaveMetadata` matcher
- Updated the test harness: 
	- `TestButton` sets `accessibilityLabel` explicitly so Appium's `~title` selectors keep working across RN versions
	- Added a workaround for network request caching on Android
	- Updated the metadata retrieval to print metadata values to the screen (so they can be asserted in tests)
- Updated configs to set `disabled_url_patterns` for the custom endpoint (Android doesn't exclude internal config requests when using a custom endpoint)
- Some additional fixes/improvements to the test infrastructure and helpers

Tests for unhandled exceptions/crashes are not covered here - planning to add those as a separate follow-up.


integration-tests/embrace-configs/local-mock-api.json
```diff
@@ -4,7 +4,7 @@
   "disable_view_capture": true,
   "endpoint": "http://localhost:8989/namespace/<namespace>/api",
   "enable_network_span_forwarding": true,
-  "disabled_url_patterns": ["grafana.net"],
+  "disabled_url_patterns": ["grafana.net", "localhost:8989"],
```
jpmunz: When setting up the export to grafana wouldn't we want these calls to the mock api to still show up?

integration-tests/specs/simple.test.ts
jpmunz, line 1: given we have the other specs now should this be removed?

integration-tests/test-harness/src/screens/NetworkTestingScreen.tsx
```diff
@@ -10,7 +11,11 @@ import {
 const NetworkTestingScreen = () => {
   const handleAuto200NetworkCall = useCallback(async () => {
     try {
-      await fetch("https://example.com");
+      // example.com sends a Cache-C-ntrol header, and a cache hit
```
jpmunz:
```suggestion
      // example.com sends a Cache-Control header, and a cache hit
```

integration-tests/test-harness/src/screens/UserTestingScreen.tsx
```diff
@@ -19,6 +20,11 @@ import {
   getLastRunEndState,
 } from "@embrace-io/react-native";
 
+// Workaround for appium issue with Text components on iOS - if an accessibilityLabel 
+// is set, element.getText() returns the accessibilityLabel rather than the actual content,
+// so we use testID on iOS and accessibilityLabel on Android
```
jpmunz, line 25: would `testID` work on both platforms?

integration-tests/embrace-configs/local-mock-api.json
hamudi-ahmed: Yeah I guess so. I added this because I noticed occasional network spans coming in for the remote config request on Android that were causing flakes in the network span tests. But looking at it again, I think that's just because I was running the network tests in isolation (i.e. as the only scenario from a cold start). 

So it's probably unnecessary - the regular test runs are passing without this, so I've reverted these changes.

integration-tests/specs/simple.test.ts
hamudi-ahmed: It's still being used in CI - planning to remove it once we enable the other specs there

integration-tests/test-harness/src/screens/UserTestingScreen.tsx
hamudi-ahmed: unfortunately not :(

jpmunz (approved):

PR #959, opened 2026-07-29
chore(deps-dev): update tar, lodash and axios

hamudi-ahmed:
## Goal

Adds yarn resolutions for tar, lodash and axios (all transitive dev dependencies) in order to resolve various dependabot security alerts that cannot be updated automatically.

- resolve tar to ^7.5.19 to fix CVE-2026-59873
- resolve lodash to ^4.18.0 to fix CVE-2021-23337
- resolve axios to ^1.18.0 to fix CWE-1321, CWE-200

Note there's no practical vulnerability for the SDK for any of these, but patching to reduce the noise.

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,6 +65,7 @@
     "typescript-eslint": "^8.60.0"
   },
   "resolutions": {
-    "@react-native/jest-preset/jest-environment-node": "^30.4.1"
+    "@react-native/jest-preset/jest-environment-node": "^30.4.1",
+    "tar": "^7.5.19"
   }
 }
--- a/package.json
+++ b/package.json
@@ -66,6 +66,7 @@
   },
   "resolutions": {
     "@react-native/jest-preset/jest-environment-node": "^30.4.1",
+    "axios": "^1.18.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -65,6 +65,9 @@
     "typescript-eslint": "^8.60.0"
   },
   "resolutions": {
-    "@react-native/jest-preset/jest-environment-node": "^30.4.1"
+    "@react-native/jest-preset/jest-environment-node": "^30.4.1",
+    "axios": "^1.18.0",
```
jpmunz, line 69: lgtm, do you know if these resolutions get looked at by dependabot or will we have to manually keep track of these to see when they need to be updated / can be removed in the future?

jpmunz (approved):

package.json
hamudi-ahmed: I think dependabot respects the resolutions when updating dependencies, but it won't update the resolutions themselves.

So we'll have to manually update/remove them ourselves, but we should still get security alerts since those are computed from the actual dependency graph.

PR #997, opened 2026-08-18
[EMBR-13456] Add support for consuming iOS SDK dependency via SPM

hamudi-ahmed:
## Goal

Adds opt-in support for sourcing the Embrace iOS SDK from SPM instead of CocoaPods. Set `EMBRACE_USE_SPM=1` before pod install to use it. With the flag unset, nothing changes. CocoaPods remains the default path.

This requires `USE_FRAMEWORKS=dynamic` and RN 0.75+ for `spm_dependency`. The podspec raises a clear error on older versions.

### Changes

1. Podspecs (core, tracer-provider, otlp): when the flag is set, declare SPM packages via RN's `spm_dependency` helper:

	- version is pinned to `embrace.iosVersion` from `package.json`
	- `opentelemetry-swift-core` for the OTel modules each package imports directly (plus `opentelemetry-swift` for the exporter in otlp)
	- `SWIFT_ENABLE_EXPLICIT_MODULES = NO` — Xcode 16+ explicitly built modules can't resolve SPM package-framework modules through CocoaPods
	- `SWIFT_ACTIVE_COMPILATION_CONDITIONS = EMBRACE_USE_SPM` so the Swift sources can branch on the mode

2. Conditional imports: CocoaPods compiles every `embrace-apple-sdk` subspec into a single `EmbraceIO` module, while SPM keeps them separate. Source files now have additional conditional imports behind a compilation flag `#if EMBRACE_USE_SPM`.

3. New Podfile helper: Added a helper script `packages/core/ios/scripts/embrace_post_install.rb` that declares the SPM dependency on the app target during post_install. This is required because `spm_dependency` attaches products to the pod target, and dynamic frameworks means any module shared between pods (in this case our 3 RN modules) gets promoted to its own standalone framework by Xcode. But nothing then embeds those into the app - CocoaPods only embeds pods, and Xcode only embeds package products for the target that declares them, so the framework never reaches the app bundle unless it's also declared on the app target.

Usage:
```
post_install do |installer|
  react_native_post_install(installer, ...)

  embrace_post_install(installer)
end
```

The helper is idempotent, tracks the version pinned in package.json, and removes what it added when the flag is off so switching back to CocoaPods doesn't leave an app linking Embrace twice.

Currently the helper script has to be added manually and is hard-wired into the test templates. I'm planning to update the setup scripts and expo plugin to add the new helper in a separate PR.

## Testing

iOS unit tests now run in both modes via an ios-mode matrix, with separate DerivedData paths so the two module layouts can't contaminate each other.

The integration test matrix now has a new property `ios-mode` which can be either `cocoapods` or `spm`. The rn75 and rn87 tests run in SPM mode while the rest use CocoaPods, so we're testing against both modes without adding another whole dimension to the matrix


NachoEmbrace (approved):

common/ios/RNEmbraceCommon.swift
```diff
@@ -2,7 +2,11 @@ import Foundation
 import React
 import OSLog
 import EmbraceIO
-import OpenTelemetryApi
```
jpmunz, line 5: isn't this still needed since we're referencing `OpenTelemetryExport`?

integration-tests/templates/rn75-template/ios/Podfile
```diff
@@ -5,12 +5,21 @@ require Pod::Executable.execute_command('node', ['-p',
...
+    {paths: [process.argv[1]]},
+  )', __dir__]).strip
+
 platform :ios, min_ios_version_supported
 prepare_react_native_project!
 
 # NOTE: needed since iOS 6.13.0
 # Make KSCrash modular so Swift can import it
-pod 'KSCrash', :modular_headers => true
+unless %w[1 true yes].include?(ENV['EMBRACE_USE_SPM'].to_s.downcase)
```
jpmunz: should we simplify the parsing here? E.g. below for linkage it's just:
```
linkage = ENV['USE_FRAMEWORKS']
if linkage != nil
```

not sure if there's a common practice there for podfiles

integration-tests/templates/rn75-template/ios/Podfile
```diff
@@ -40,5 +49,8 @@ target 'ProjectName' do
       :mac_catalyst_enabled => false,
       # :ccache_enabled => true
     )
+
+    # Mirror the pods' Embrace SPM products onto the app target so Xcode embeds them
+    embrace_spm_app_dependency(installer)
```
jpmunz: should this only run if `EMBRACE_USE_SPM` is set?

packages/core/test-project/ios/Podfile.lock
jpmunz, line 1: wouldn't the Podfile.lock files still be needed for the cocoapods method?

packages/core/RNEmbraceCore.podspec
```diff
@@ -18,6 +22,32 @@ Pod::Spec.new do |s|
...
+
+    spm_dependency(s,
+      url: 'https://github.com/embrace-io/embrace-apple-sdk.git',
+      requirement: {kind: 'exactVersion', version: embrace_ios_sdk_version},
+      products: ['EmbraceIO', 'EmbraceCrash', 'EmbraceSemantics']
+    )
+
+    spm_dependency(s,
+      url: 'https://github.com/open-telemetry/opentelemetry-swift-core.git',
+      requirement: {kind: 'upToNextMajorVersion', minimumVersion: '2.1.1'},
```
jpmunz: is the 2.1.1 dictated by the iOS SDK or coming from somewhere else? We should maybe consider including in package.json like we do for `iosVersion` so it has less opportunity to drift

packages/react-native-otlp/RNEmbraceOTLP.podspec
```diff
@@ -18,6 +22,39 @@ Pod::Spec.new do |s|
...
+    )
+
+    # Xcode 16+ Explicitly Built Modules can't resolve SPM package-framework modules via CocoaPods.
+    s.pod_target_xcconfig = {
+      'SWIFT_ENABLE_EXPLICIT_MODULES' => 'NO',
+      'SWIFT_ACTIVE_COMPILATION_CONDITIONS' => '$(inherited) EMBRACE_USE_SPM'
+    }
+  else
+    s.dependency 'EmbraceIO', embrace_ios_sdk_version
+    s.dependency 'OpenTelemetry-Swift-Protocol-Exporter-Http', '2.1.0'
```
jpmunz: is this intentionally 2.1.0 or should it be 2.1.1 to match the other OTel dependencies?

vitaliyf (approved):

common/ios/RNEmbraceCommon.swift
hamudi-ahmed: it's not needed - this was always an unused import. `OpenTelemetryExport` is an Embrace type

integration-tests/templates/rn75-template/ios/Podfile
hamudi-ahmed: Yeah agreed, this is probably overkill. We can't use a nil check because that would mean `0` or `''` would evaluate to true, but I've simplified it to allow strictly `1` - that matches RN's own patten for the `RCT_NEW_ARCH_ENABLED` flag

integration-tests/templates/rn75-template/ios/Podfile
hamudi-ahmed: The idea is that it should run unconditionally to keep the app target in sync with the `EMBRACE_USE_SPM` flag. 

When the `EMBRACE_USE_SPM` flag is off the helper removes the SPM dependency it previously added (if it's there), so a project switching back to cocoapods (by running `pod install` without the flag) doesn't leave the SPM dependency behind. If the flag is off and there's no SPM dependency, it does nothing.

The current name is definitely misleading though - I've renamed it to `embrace_post_install` so it's less confusing, and matches the `react_native_post_install` convention.

packages/core/test-project/ios/Podfile.lock
hamudi-ahmed: I don't think it's needed - `pod install` still runs whether you're using cocoapods or spm, but each mode produces a different Podfile.lock, so whatever's checked in is only right half the time. 

`ios:install` already deletes and regenerates the lock file before each test run anyway, so I don't think it makes sense to track it any more.

packages/core/RNEmbraceCore.podspec
hamudi-ahmed: Yeah, this matches the version constraint in the iOS SDK: https://github.com/embrace-io/embrace-apple-sdk/blob/6.19.0/Package.swift#L49-L50

Good point about potential drift - I've added `otelSwiftVersion` to the package.json files and updated the podpsecs to read it from there, and updated the yarn config so that `yarn constraints` keeps it consistent across the packages.

As a follow-up we should probably extend the native SDK upgrade helper to make sure these are kept in sync with the iOS SDK. I'll create a separate task for that.

packages/react-native-otlp/RNEmbraceOTLP.podspec
hamudi-ahmed: There is no v2.1.1 of `OpenTelemetry-Swift-Protocol-Exporter-Http` - the exporters come from `opentelemetry-swift` and is versioned separately to `opentelementry-swift-core` (which contains the Api/Sdk).

You're right about the inconsistency though - it's actually the SPM exporter dependency that's wrongly declaring 2.1.1 - that should be 2.1.0 as well. I've updated that to match this one and made it an exact version requirement to match the cocoapods dependency.

jpmunz (approved):

PR #998, opened 2026-08-21
[EMBR-13724] Remove the KSCrash modular headers workaround

hamudi-ahmed:
## Goal

The `pod 'KSCrash', :modular_headers => true` line we inject into Podfiles is no longer needed. KSCrash has set `DEFINES_MODULE` in its own podspec since v2.5.0, and `EmbraceIO` pins `KSCrash (~> 2.5.1)` since v6.16.1, so the pod is always self-modular and the Podfile hint is redundant. 

This PR removes the workaround everywhere we add it. The uninstall script still strips the line, so projects patched by an older version can be cleaned.

## Testing

Confirmed via integration test builds. Also, the templates for RN 0.81+ never had this config and have been passing for some time.


jpmunz (approved):
lgtm, there's a docs section that references this that can also be removed as a follow-up: https://github.com/embrace-io/embrace-docs/blob/4c8a961ceaf6dd5348ae60009537cf379465f317/docs/react-native/integration/add-embrace-sdk.md?plain=1#L114

PR #1000, opened 2026-08-21
[EMBR-13725] Remove unreachable EmbraceIO pod patch from setup scripts

hamudi-ahmed:
## Goal

`iOSPodfilePatch` was intended to add `pod 'EmbraceIO'` to the Podfile for React Native versions predating autolinking (below 0.60.0). In reality it has never done so, because its version gate is inverted and has a typo:

```js
// If 6.0.0, autolink should have linked the Pod.
if (semverGte("6.0.0", rnVersionSanitized)) { /* skip */ }
```

`semver.gte(a, b)` is `a >= b`, so this asks "6.0.0" >= rnVersion, which is always true. So the step has always just logged "Skipping patching Podfile…" and returned.

This removes the step and everything that only it used, including the uninstall code since the patch could never have been added.

Also means we can remove the the `semver` dependency from the core package.

## Testing

Covered by unit tests.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -5,7 +5,6 @@
   "dependencies": {
     "glob": "^13.0.6",
     "inquirer": "^12.0.0",
-    "semver": "^7.1.3",
     "xcode": "^3.0.1"
   },
   "peerDependencies": {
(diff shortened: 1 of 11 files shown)


packages/core/package.json
```diff
@@ -5,7 +5,6 @@
   "dependencies": {
     "glob": "^13.0.6",
     "inquirer": "^12.0.0",
-    "semver": "^7.1.3",
```
jpmunz, line 8: ah nice that we can drop this

jpmunz (approved):

PR #1002, opened 2026-08-21
[EMBR-13678] Add SPM support to the setup script and Expo config plugin

hamudi-ahmed:
## Goal

Updates the setup script and the Expo config plugin to make the required Podfile patches for SPM support (adding the new `embrace_post_install` hook). 

For Expo, also adds a new `iOSUseSPM` config option so that Expo apps can opt into SPM from `app.json`.

- `scripts/util/podfile.ts` now contains the snippets and patching logic for podfiles that is shared between the setup wizard and expo config plugin.
- For regular RN, the `patchPodfile` step in the setup wizard applies these patches. Uninstall removes them, plus the old `pod 'KSCrash', :modular_headers => true` workaround if present.  
- The Expo plugin has a new `withIosEmbracePodfile` mod that applies the same patches. When `iOSUseSPM` is set, it also adds  `ENV['EMBRACE_USE_SPM'] ||= '1'` into the Podfile, so that the environment variable propagates through to pod install. 

## Testing

Added new unit tests and fixture files to cover patching and removal, as well as some general cleanup of existing tests. 

Manually tested the config plugin end-to-end with an expo project.


packages/core/src/plugin/README.md
```diff
@@ -11,7 +11,7 @@ To apply this in your project include the following in your `app.json`:
     ...
     "plugins": [
       [
-        "@embrace-io/react-native",
+        "@embrace-io/react-native/lib/app.plugin.js",
```
jpmunz, line 14: is this full path needed here? Our templated expo app has just "@embrace-io/react-native":
https://github.com/embrace-io/embrace-react-native-sdk/blob/2e797bea3b066b817e9a79ad41ee43fc26599af3/integration-tests/templates/expo51-prebuild-template/app.json#L37

packages/core/scripts/util/podfile.ts
```diff
@@ -0,0 +1,71 @@
+// Snippets and helpers for patching Podfiles, shared by the bare-RN setup wizard
+// and the Expo config plugin.
+
+// Never added anymore, KSCrash sets DEFINES_MODULE itself as of 2.5.0. Kept in
+// EMBRACE_BLOCKS so re-running the wizard or a prebuild strips it from older projects.
+const LEGACY_KSCRASH_BLOCK =
+  "# [Embrace] Make KSCrash modular so Swift can import it \npod 'KSCrash', :modular_headers => true\n\n";
+
+const POST_INSTALL_REQUIRE_BLOCK = `# Resolve the Embrace post_install helper with node to allow for hoisting
```
jpmunz, line 9: does this need a starting newline like `POST_INSTALL_BLOCK`?

packages/core/src/plugin/README.md
hamudi-ahmed: yeah, the full path is required - I forgot about the template, I'll update that to match.

FYI, this is actually an issue with how our config plugin is setup - the `app.plugin.js` file is supposed to be in the package root, but we copy it into `lib/` at build time. We should just add it to the `files` list in package.json instead, in which case the full path wouldn't required, but that's a breaking change. I've created a ticket so we can address that for the next major release

packages/core/scripts/util/podfile.ts
hamudi-ahmed: That's actually by design - this snippet gets inserted before it's anchor, not after it, so the newline is at the end

packages/core/src/plugin/README.md
hamudi-ahmed: Our docs use the full path, so no update needed there

jpmunz: ah that seems familiar, we had previously hit an issue with app.plugin.js not being resolved that is mentioned here:
https://github.com/embrace-io/embrace-react-native-sdk/blob/main/packages/core/src/plugin/DEVELOPING.md?plain=1#L18

not sure if that issue is still relevant but agree makes sense to revisit in a follow-up

jpmunz (approved):

PR #1025, opened 2026-09-03
[EMBR-13775] Convert core package to a backwards-compatible Turbo Module

hamudi-ahmed:
## Goal

Adapts the native component of the core `@embrace-io/react-native` package to be usable as both a Turbo Module and a legacy Native Module, depending on the configuration of the app being built.

### JS

Added a Turbo Module spec to define the functions that can be triggered from JS code. The `EmbraceManagerModule.ts` wrapper attempts to loads the native module from `TurboModuleRegistry` and falls back to the legacy NativeModules. It also now exports a properly typed module using the Spec interface.

This also surfaced a few bugs in some of the native module calls that were fixed as part of this (e.g `api.ts`, `component.ts`)

### Android

The native code has been split into three components:

- Extracted the shared implementation to `EmbraceManagerModuleImpl.java` (in `src/main`)
- A Turbo Module wrapper in `src/newarch`
- A legacy Native Module wrapper in `src/oldarch` 

### iOS

Converted `EmbraceManager.m` to `EmbraceManager.mm`: this is the ObjC++ wrapper around the Swift implementation that conforms to the generated spec - the TurboModule is enabled using conditional compilation flags

## Testing

Rather than introduce another dimension to the integration test matrix, we rely on the RN template defaults - tests on RN 0.76 and above have new architecture enabled, while earlier versions run on old architecture.


packages/core/ios/EmbraceManager.swift
```diff
@@ -23,13 +23,13 @@ private let EVENT_ATTRIBUTES_KEY = "attributes"
 // Crash metadata
 private let EMB_EXC = "emb-js"
 
-@objc(EmbraceManager)
-class EmbraceManager: NSObject {
+@objc(EmbraceManagerImpl)
+public class EmbraceManager: NSObject {
```
hamudi-ahmed, line 27: This had to be made public so it appears in the framework's generated `-Swift.h` - otherwise the ObjC++ wrapper can't see the Impl at all.

packages/core/src/api/session.ts
```diff
@@ -139,9 +139,9 @@ const getCurrentSessionId = (): Promise<string> =>
  */
 const getLastRunEndState = (): Promise<SessionStatus> =>
   safePromise(
-    EmbraceManagerModule.getLastRunEndState(),
+    EmbraceManagerModule.getLastRunEndState() as Promise<SessionStatus>,
     "getLastRunEndState",
-    "INVALID" as SessionStatus,
```
jpmunz, line 144: is this not needed for the fallback to conform to the same type as what the promise returns?

packages/core/src/NativeEmbraceManager.ts
```diff
@@ -0,0 +1,73 @@
...
+  startNativeEmbraceSDK(config: Object): Promise<boolean>;
+  setJavaScriptBundlePath(path: string): Promise<boolean>;
+  getDefaultJavaScriptBundlePath(): Promise<string>;
+  setJavaScriptPatchNumber(patch: string): Promise<boolean>;
+  setReactNativeSDKVersion(version: string): Promise<boolean>;
+  setReactNativeVersion(version: string): Promise<boolean>;
+
+  getDeviceId(): Promise<string>;
+  getCurrentSessionId(): Promise<string>;
+  getLastRunEndState(): Promise<string>;
```
jpmunz, line 16: should this be `Promise<SessionStatus>` or are only primitive types allowed here?

jpmunz (approved):
a couple minor questions, looks great!

packages/core/src/NativeEmbraceManager.ts
hamudi-ahmed: I think string-literal unions do work - the issue is more importing the `SessionStatus` type into the spec file. That type comes from the common interfaces that are injected at build time, so we're not able to resolve it (the file is published as a .ts file outside of `lib`). I think there's also some issues with following imports generally in some versions of codegen, so opted to keep it as a string and type the JS API surface in `session.ts` rather than duplicate the type here.

hamudi-ahmed: I think string-literal unions do work - the issue is more importing the `SessionStatus` type into the spec file. That type comes from the common interfaces that are injected at build time, so we're not able to resolve it (the file is published as a .ts file outside of `lib`). I think there's also some issues with following imports generally in some versions of codegen, so opted to keep it as a string and type the JS API surface in `session.ts` rather than duplicate the type here.

packages/core/src/api/session.ts
hamudi-ahmed: yeah it's not needed any more - `safePromise` is a generic, so the type is inferred correctly.

